PR #15, opened 2021-10-22
Possible fix of workflow syntax for cmake --build config

nprorekhin:
Hi,

As far as I understood from Actions page, the last windows builds were unsuccessful. I am not sure, but it look like this is caused by environmental variable passed in `cmake --build . --config $BUILD_TYPE` . I used a different syntax in my other project, so maybe this `cmake --build . --config ${{env.BUILD_TYPE}} `will do the trick? 

Or I am wrong here?


ccifra:
This used to work and I pulled it from a template.  I am no export on actions.  Let's give this a try.

PR #28, opened 2021-11-08
Relinked controls and VIs linked to vi.lib

KarthikAbiram:
There were VIs and controls linked to a release lvlib (gprc-lvsupport-release.lvlib) that is expected at vi.lib. I assume they need to be linked within the source code folder as below. @ccifra 

image


navinsubramani:
@KarthikAbiram  The last commit has refactored the code by splitting the lv-support and templates into different project. So, the development approach is as mentioned in the md file - https://github.com/ni/grpc-labview/blob/master/docs/developing%20and%20building%20LabVIEW%20libraries.md

templates source should be dependent on the "grpc-lvsupport-release", but I see that we are yet to release the package, so you might have challenges to link it to the library that you dont have. Please see if you can follow the instruction and update the solution? ping me if you have questions.

KarthikAbiram:
Sure - in that case, I will go ahead and close this pull request.

Issue #42, opened 2021-11-20
New Modular Service Template


navinsubramani:
Some of the current limitations of the previous templates are,
1. Template combines RPC messages and RPC methods into a single class (but proto can have many services per message)
2. The server creation and close is coupled with the service methods and messages functions
3. Only one service can run on a port number

The requirement for the new template is to,
1. Modularize RPC messages, RPC service(s) and gRPC server/servicer functionality
2. Enable user to add more than one service class to the library and run them on same port
3. Create API and parent classes and abstract the common functionality out of the specific library

Some design note:
![image](https://user-images.githubusercontent.com/17029551/142708029-1de8abef-862b-4cfb-83e4-65ba201aaa99.png)

![image](https://user-images.githubusercontent.com/17029551/142708061-2f6e2e55-0917-4db7-b08e-a24e43dccea8.png)

![image](https://user-images.githubusercontent.com/17029551/142708071-38415103-36e8-4ab7-8c20-513de90de17d.png)

![image](https://user-images.githubusercontent.com/17029551/142708082-46452d5d-c979-402d-aba1-d85e6a2effec.png)

![image](https://user-images.githubusercontent.com/17029551/142708098-ba8254ce-145d-463e-b33c-b85eb0ca018f.png)

chrisb2244:
Obviously I'm new to all of this, so feel free to ignore these thoughts, but here they are anyway:

- Most LabVIEW libraries I've seen don't use the "i\<Feature\>" naming, even for interface-like/abstract-ish classes. Perhaps a name like "gRPC Service" (or "RPC Service"?) would be more "LabVIEW-like"?
- Your proto file has "SignalGeneratorService2" (no space), whilst the class became "SignalGeneratorService 2.lvclass". Was this just an accident in the example (I guess from the comment on #17 that this was created manually and not scripted?) or is this an intentional naming choice?

Otherwise seems very nice to me. Will download the code in the PR and look at implementation too, but as I said please don't think too much of my thoughts!

navinsubramani:
@chrisb2244  Since, this is hand created example, I created the proto and code separately. You are right #2 is a mistake. The name of the LabVIEW class will be the same as the service defined in the proto.

#1 good thought. I started that as an interface class but I don't think it's an interface. We could change that something like ServiceBase, as it has some base implementation in them. Good point!

Issue #16, opened 2021-10-26
Limitation with auto code generation when the package name has 'dot'


navinsubramani:
When the proto file package name has 'dot' in it. For example: Service.V1 where the version of the service is represented as part of the package name, and the script is used to generate the code. Code generated doesnt have any methods in it, and no error are generated.

ccifra:
Issue is in the "Translate gRPC Method FullName.vi" VI which is not properly converting the method name and assuming that there are only 2 dots in the name.

navinsubramani:
@ccifra  Have added the change on top of the existing pull request

Issue #4, opened 2021-05-04
Missing DLLs in gRPC Scripting Tools Project


sreeja2018:
Hi Team,

Greetings!!!

The DLL 'labview_grpc_generator.dll' is not available in the repo for all platforms (Win32, Win64, Linux, LinuxRT)
![image](https://user-images.githubusercontent.com/35146913/116982723-c86d6480-ace6-11eb-9789-1de70ce6e513.png)

Kindly assist in uploading them to the repo in the earliest.

Thanks

sreeja2018:
Found them in previous versions.
Thanks

sreeja2018:
The previous versions had the dll for Windows only
Please add the dlls of all platforms (Win32, Win64 and Linux) to repo.

sreeja2018:
Hi Team,
Previously, I had downloaded the source code (zip) where the function libraries were missing.
I found all the function libraries for all the platforms in the package grpc-labview.zip in Version 0.2.3.
Sorry for the inconvenience.
Thanks

Issue #1, opened 2021-01-13
Only first element of a gRPC server stream is transmitted in QueryServer_Event_Fire.vi


nprorekhin:
Hi,

I've discovered one thing when working with the example. 

It seems like _example\VIs\QueryServer_Event_Fire.vi_ transmits only the first element, however it is a gRPC stream and we should see different messages arriving to the client.

Steps to reproduce:
1. Modify the shipping example ExampleQueryServer.vi. Open Timeout case and add some dynamic information to the "Hear Beat" message. I did it with server uptime.
![gRPC-stream](https://user-images.githubusercontent.com/53332180/104469770-2a2f8480-55b9-11eb-920e-3c10289b7658.png)
2. Start _example_client.exe_
3. We do receive 10 stream messages but they are the same:
```
Server uptime: 15.393039
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server Event: Heart Beat, uptime 16.397 seconds
Server notifications complete
```

Am I missing something or is gRPC server streaming is not yet fully implemented?

ccifra:
Fixed with 58e404ab8ed0578b47eebe31acef6613cc10676a.

PR #43, opened 2021-11-20
New Service Template

navinsubramani:
#42  The issue is mentioned here.

Note, the development approach is similar to how the "current template" and "gRPC LabVIEW support" is linked and developed. Below is the dependency diagram.
image

Please build and install the "LabVIEW gRPC Servicer" to use the "Client Server Template New". The package that is linked here (rename the extension from .txt to .vop and install the package or build it using the build spec in the source)
[ni_lib_labview_grpc_servicer-0.3.0.5.txt](https://github.com/ni/grpc-labview/files/7573589/ni_lib_labview_grpc_servicer-0.3.0.5.txt)
,


navinsubramani:
@ccifra  Committed the template components alone. Once this is reviewed and signed off, we can create an issue for the scripting tool that generates the template from one or more proto files.

navinsubramani:
This issue is mentioned here #16 

Before this change:

When the proto file package name has a "dot" in it. For example: Package.V1.

During creation of server code, we get the server code created with issues in RPC methods
image

During creation of client code, we get an below error and client code is not created
image

This issue is fixed by doing changes in two VIs:

"Translate gRPC Method FullName" VI is changed to consider a "dot" in the package name.

From:
image

To: 
image

"Create LV RPC Method Name" VI is changed to consider a package name with dot and replace it with underscore to create a Method Name.

From: 
image

To: 
image

Test Cases:
Tested both code generation of server and client code, and basic functionality of server and client for all the below cases,
1. Proto file with no package name
2. Proto file with package name and DOT
3. Proto file with package name and no DOT

ccifra:
Some comments on the overall structure:

End RPC Call.vi is under the RPC Messages folder.  Is there a reason for that?  It is only needed when implementing a server and I think it should be part of the server class.  I am not sure we even need it since it is just a simple wrapper around the end call VI in the LV Support library.

For the messages, do we need to add _Data to the typedef of the message or can it be just be package_templatemessage.ctl

Given the path length issues do we need to have a folder per service method?  I think it would be better to just have a single folder for all methods with a messaging and registration folder under that, but all of the method VIs would be in the same folder.

What is the purpose of "RPC Method package_service_method logic.vi"

navinsubramani:
@ccifra 

#1 I am good with removing it, since like you mentioned it is a wrapper. All the messages contain the "set" API which closes the RPC call themselves. If we are planning to make the "end call" VI public for developers, then we can remove them.

#2 I dont see any naming spacing concerns for the message to have a _Data with it. It came from legacy template. I will remove that as well.

#3 I have them flat in file explorer, and only on the project made some virtual folders. I will probably have only a virtual folder for each method and not have many nest under it. But the files in disk are flat under RPC methods
image

#4 Given the business logic are written in the event structure, and the logic should be repliacted in Sync & Async engine and it is upto developers to create VIs to modularize them - I have created a VI for it. This VI will be dropped in the event structure of sync and Async for their respective events. So, developers need not go to sync or async engines, rather just add their service logic in this VI and both sync/async will work well.  Also, when there are changes to service in proto file, the events will break but if the logic is in this VI, we could avoid disturbing the logic.

This is the idea. Please let me know your thoughts.

navinsubramani:
@ccifra I was recommending folks to create a VI under event structure for the logic, and this VI will help that. But maybe we need not have this for version 1, and look at it in future if this comes up. Less development now :)

Its easy to tell people to code here,
image

Rather than code here,
image

ccifra:
That does not work as well for the streaming case or when the method changes what is being stream.  I think we should think about this aspect a little more before moving forward.

navinsubramani:
Sure @ccifra . I was thinking about streaming and imagining will that be a blocking call or Asynchronous implementation. I guess if its in event structure case it will be a blocking call (even in VI) for other events. The logic should be implemented outside the event case for server streaming, if it should allow for other event calls meanwhile. I guess I will wait to look at the APIs and decide. For now, Will not build any placeholders for logic - It will be open.

navinsubramani:
@ccifra Made changes mentioned above.

ccifra:
This was resolved by PR #46 which included these changes.

PR #46, opened 2021-12-11
Service Script Utility for the new template

navinsubramani:
- Create a scripting utility for the service


navinsubramani:
- FIxed the "dot" issue
- New template for service : Can hold any service across different protos (still the parser will parse only one service per proto, but when that fixed internally this tool will work accordingly)
- Scripting Tool for the new service template
- Path issue is considered: The files are NOT created in deep paths

ccifra (approved):

PR #49, opened 2021-12-12
New Client Template

navinsubramani:
- Created New Client Template


ccifra:
Template looks good.

ccifra (approved):

navinsubramani:
@ccifra  Please review the new client tool

- The client will not work if we need to update the client code with a changed proto. Given, client will not have use generated code, users can recreate client from scratch as a temp fix
- I have tested the "Server Stream" and "Unary" - There are issues during run with "Client Stream". I am not sure what are the use cases for "Bidirectional". Please use this and test the solution for these two cases and send me a example if possible.
- Fixed the issue that is reported recently - when script is rerun with a proto containing new message.

ccifra (approved):

PR #74, opened 2022-03-25
Client Template Timeouts

BShermanator:
Added an optional timeout input to the Client Methods. Based on the resources below, I believe the additions I made fall under the definition of a "timeout" and not a "deadline". 

I left the unit in milliseconds since I do not know what a reasonable timeout value will be for each application. In my project leveraging this repo, I have some functions that need a 30 second timeout, but many examples I find online use 100 milliseconds. 

I copied the updates to my development machine and ensured the older scripting tool still works. The optional input was nice because it does not break all of the existing code I have so far. 

Resources:
- https://grpc.io/blog/deadlines/
- https://mchesnavsky.tech/grpc-timeouts-deadlines/

VIM Change:
image

Template VI Change:
image


ccifra:
I think we should incorporate the timeout into the actual gRPC call instead of just making it a timeout on the occurrence.  This way the allocated resources on the client side of the call can be cleaned up if there is a timeout.  Also, the server will be able to check for a timeout if it wants to.

I will update the C code to take in the timeout and pass it along to the call.  Then we can update this PR to pass the timeout to the call.

BShermanator:
Sounds great! Let me I can make those updates.

ccifra:
Fixed by #98

PR #78, opened 2022-04-24
Change the Reentrancy of start related VIs to "shared clone reentrant execution"

gaoruhao:
Change the Reentrancy of the following VIs to "shared clone reentrant execution". By doing so, the gRPC service class can run in multiple instances parallelly. The example VIs are updated as well.

These VIs are:
- Start.vi
- Start Async.vi
- Start Sync.vi


gaoruhao:
@ccifra Todd and I are using this lib to build ADAS project. We discovered the needs of re-entrant VI recently when trying to start multiple instances of the grpc service.

ccifra (approved):

PR #94, opened 2022-06-16
Remove infinite loop that can occur

pakdev:
Without this change, LabVIEW hangs in `Create Client.vi > CompleteMetadataRegistration.vi > CompleteMetadataRegistration` if messages are nested deeper than 2 levels. For example, this works...

```
message Ego {
  int32 x = 1;
}

message GoProResponse {
  Ego ego = 1;
}
```

But this doesn't...

```
message Quantity {
  float value = 1;
  string units = 2;
}

message Ego {
  Quantity x = 1;
}

message GoProResponse {
  Ego ego = 1;
}
```

Debugging the dll points to the removed loop as the culprit, and it doesn't appear to do anything. When I removed it, built, and streamed back my deeply nested message response, everything worked.

### DIFF
--- a/src/message_element_metadata_owner.cc
+++ b/src/message_element_metadata_owner.cc
@@ -78,7 +78,7 @@ namespace grpc_labview {
         {
             return 0;
         }
-        auto multiple = ClusterElementSize(type, repeated);    
+        auto multiple = ClusterElementSize(type, repeated);
         int remainder = abs(clusterOffset) % multiple;
         if (remainder == 0)
         {
@@ -95,7 +95,7 @@ namespace grpc_labview {
     void MessageElementMetadataOwner::RegisterMetadata(std::shared_ptr<MessageMetadata> requestMetadata)
     {
         std::lock_guard<std::mutex> lock(_mutex);
-        
+
         _registeredMessageMetadata.insert({requestMetadata->messageName, requestMetadata});
     }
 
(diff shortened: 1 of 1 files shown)


src/message_element_metadata_owner.cc
```diff
@@ -118,26 +118,20 @@ namespace grpc_labview {
...
+        }
         int clusterOffset = 0;
         for (auto element: metadata->_elements)
         {
             auto elementType = element->type;
             auto nestedElement = element;
-            while (elementType == LVMessageMetadataType::MessageValue)
-            {
-                auto nestedMetadata = FindMetadata(element->embeddedMessageName);
-                auto nestedElement = nestedMetadata->_elements.front();
```
ccifra, line 130: I think we need to keep the while loop.  What this code is doing is computing cluster element memory offset (the padding of the elements to keep byte alignment).  When there is a nested message the alignment is based on the first element of the message that itself is not a message.

I think the bug is that we are always calling FindMetadata passing in element->embeddedMessageName and not the changing element to be nestedElement for the next iteration of the while loop.

I would chnange is so that we create a currentElement initialized to element and use that for FindMetadata and after each iteration we change currentElement to equal nestedElement.

ccifra (changes requested):
See comments in code changes.

src/message_element_metadata_owner.cc
pakdev: Would `currentElement` then replace usages of `element` following the loop? Otherwise, what is the loop mutating?

pakdev: Also, aren't we already computing cluster offsets for nested messages in lines 135-151 (old), 129-145 (new). Notice the call is recursive

pakdev:
> See comments in code changes.

See reply to your comment

ni-sujain:
this file has been changed significantly, please re-test and create an issue if its there, closing this PR.

Issue #76, opened 2022-03-31
Need Timeout in Client Calls


becega:
When doing calls to the server, if the function is partially implemented (there is an event handle structure but it doesn't return the value). The client waits forever, hanging the application. We need to specify a timeout for how long we can wait for the query to complete.

BShermanator:
I have an open PR already for your suggestions. I ran into the same issue. 

https://github.com/ni/grpc-labview/pull/74

ni-sujain:
fixed by #98

PR #98, opened 2022-07-20
Implementation of client timeouts

shivaprasad-basavaraj:
Fix for #76 

This PR contains changes to implement timeouts on the grpc client. The timeouts have been implemented using grpc deadlines for both the unary and streaming calls. This would mean that the timeouts in the case of streaming calls would be for the whole of the streaming call and not individual read/writes of the streaming call. 

I have also updated the LV code to pass in the timeout and the template VIs to include the timeout option. The timeouts in these VIs are optional and are -1 by default which means use the default timeout that grpc provides.

Changes to the VIs have been captured below.

- Client Unary Call.vim
image
- Client Begin Server Streaming Call.vim
image
- Client Begin Client Streaming Call.vi
image
- Client Begin Bidirectional Streaming Call.vi
image
- RPC Method Template.vi
image
- Unary UnaryTemplate.vi
image
- Begin ServerStreamTemplate.vi
image
- Begin ClientStreamTemplate.vi
image
- Begin BidirectionalStreamTemplate.vi
image


BShermanator:
This is great news! Now I need to figure out how to cancel PR 74 I had submitted. @ccifra - are you able to remove/abandon the PR I created?

ni-sujain (approved):

BShermanator:
I have a suggestion for the timeout control naming for the LabVIEW side. Units are typically shown in parenthesis: timeout (ms)

ccifra (approved):

shivaprasad-basavaraj:
> I have a suggestion for the timeout control naming for the LabVIEW side. Units are typically shown in parenthesis: timeout (ms)

@BShermanator Thank you for the feedback. I have updated the naming of the timeout parameter.

PR #99, opened 2022-07-27
Changes to cleanup grpc connections on aborting VIs

shivaprasad-basavaraj:
Changes made in this PR are:

- Added a new LV callback for RTSetCleanupProc. This would be used to setup cleanup procs which run when a VI goes to Idle.
- Changes in the grpc_client to register a cleanup proc when we create a grpc client and de-register it when we delete the client object.
- Now the grpc_client also keeps track of all the client calls that are active. This was needed since we need to cancel all these before we close the client connection.
- Changes in the grpc_server to register cleanup proc when we create the grpc server object. Now we pass a deadline of system_clock::now() to the server shutdown call so that we also shutdown any ongoing RPC calls immediately.

### DIFF
--- a/src/grpc_client.cc
+++ b/src/grpc_client.cc
@@ -152,6 +152,8 @@ namespace grpc_labview
     }
 }
 
+int32_t ClientCleanUpProc(grpc_labview::gRPCid* clientId);
+
 //---------------------------------------------------------------------
 //---------------------------------------------------------------------
 LIBRARY_EXPORT int32_t CreateClient(const char* address, const char* certificatePath, grpc_labview::gRPCid** clientId)
@@ -161,6 +163,7 @@ LIBRARY_EXPORT int32_t CreateClient(const char* address, const char* certificate
     auto client = new grpc_labview::LabVIEWgRPCClient();
     client->Connect(address, certificatePath);
     *clientId = client;
+    grpc_labview::RegisterCleanupProc(ClientCleanUpProc, *clientId);
     return 0;
 }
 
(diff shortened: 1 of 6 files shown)


src/grpc_client.cc
```diff
@@ -173,10 +176,26 @@ LIBRARY_EXPORT int32_t CloseClient(grpc_labview::gRPCid* clientId)
...
 
+int32_t ClientCleanUpProc(grpc_labview::gRPCid* clientId)
+{
+    auto client = clientId->CastTo<grpc_labview::LabVIEWgRPCClient>();
+    if (!client)
+    {
+        return -1;
+    }
+
+    for (auto activeclientCall = client->ActiveClientCalls.begin(); activeclientCall != client->ActiveClientCalls.end(); activeclientCall++)
```
ni-sujain, line 192: minor activeclientCall->activeClientCall

src/grpc_client.cc
```diff
@@ -278,12 +300,15 @@ LIBRARY_EXPORT int32_t ClientBeginClientStreamingCall(grpc_labview::gRPCid* clie
 
     auto clientCall = new grpc_labview::ClientStreamingClientCall(timeoutMs);
     *callId = clientCall;
+    clientCall->_client = client;
```
ni-sujain, line 303: question, the client was set for unary call already in the existing code, but here we are setting that for streaming call. Is that a bug with existing code.

src/lv_interop.h
```diff
@@ -52,6 +52,7 @@ namespace grpc_labview
     typedef int32_t MagicCookie;
     typedef MagicCookie LVRefNum;
     typedef MagicCookie LVUserEventRef;
+    typedef int32_t(*CleanupProcPtr)(gRPCid* data);
```
ni-sujain, line 55: should we call it id here instead of data? is there some convention for declaring cleanuprocs?

src/lv_interop.cc
```diff
@@ -9,17 +9,22 @@
 #include <dlfcn.h>
 #endif
 
+static int kCleanOnRemove = 0;
```
ni-sujain, line 12: lets add some brief comments here to explain there purpose

ni-sujain (approved):
looks good

src/grpc_client.cc
shivaprasad-basavaraj: Yes. We weren't setting the _client property on the ClientCall for the streaming calls.

src/grpc_client.cc
shivaprasad-basavaraj: Fixed.

src/lv_interop.h
shivaprasad-basavaraj: Updated.

src/lv_interop.cc
shivaprasad-basavaraj: Done.

ni-sujain:
@ccifra I am merging this change. If you have any concerns/reviews on this please let us know.

PR #107, opened 2022-08-09
Changes to automate building of VIPB files

shivaprasad-basavaraj:
- Created a python script to automate the building of VIPB files. This script copies the C++ binaries to the correct locations and then calls into LabVIEW CLI to run a BuildGRPCPackages.vi. This VI makes use of the VIPM APIs to build the VIPB files in the correct order.
- Updated the Building.md file to add details of how to use the script to automate the building of VIPB files.


docs/Building.md
```diff
@@ -140,3 +142,52 @@ Build Release
...
+```
+
+The accepted vaues for BUILD_TARGET are:
+- Win32
+- Win64
+- All
+
+If you choose to build for Win32 or Win64 target you can specify the *--buildcpp* option to indicate that you want to build the C++ binaries first and then use them to build VIP files. If you don't specify this option it is assumed that the required binaries are already built accoriding to the steps described above.
+
+If you choose to build "All" target then you need to specify a folder where we can find the pre built binaries for all the supported targets using the *--pathToBinaries** option. We expect the folder specified to have the following structure.
```
ni-sujain, line 178: Lets just make it clear here that with --All , even if one gives --buildcpp it just ignores it and need --pathToBinaries

ni-sujain (approved):
This is great as a first step towards automation

docs/Building.md
shivaprasad-basavaraj: Done.

shivaprasad-basavaraj:
@ccifra With this change the automated build scripts only the `Client Server Support New` package and not the old package since I wasn't sure we should continue to maintain it. If the old client server scripting infrastructure is not to be used anymore should we just go ahead and delete it and stop shipping the corresponding vip file with our releases.

PR #108, opened 2022-08-10
Fixed Query_server project

### LINKED ISSUE #96: Example `Query Server` does not work due to incorrect server DLL path
The `Create Server` VI fails as there is no DLL at the path expected. The VI to generate the path is expecting a directort called `dlls` and this is not present.

image

Note: the `helloworld` example uses the `grpc-lvsupport-release.lvlib` and works fine whereas the Query Server example uses `grpc-lvsupport.lvlib` which seems to be a bit different.

abahetik:
In this PR I have made changed to the Query_Server project inorder to fix #96 :
1. This project didnot use a template server . It had created it own implementation. So,
 a. I updated the server using its proto file through the grpc code generator.
 b. I implemented its `Start Sync.vi` for the example based on the proto file.
2. I removed any unrequired dependencies
3. I changed the `gprc-library-release` dependency to `grpc-library-release`


ni-sujain:
Lets do the error wire changes as we discussed

abahetik:
@ni-sujain I have implemented the error handling for the project

PR #118, opened 2022-08-22
Exposing ClientContext's TryCancel() to the LabVIEW API to prevent hanging on Finish()

### LINKED ISSUE #116: "Client Complete Streaming Call.vi" hangs if there are data items left to read
When trying to close the client stream, it hangs if there are still stream items to read out. I have tried adding a while loop to read everything out and ensuring the server sends the last data item (with false set for streaming), but if the server happens to send another items after reading all data but before the client can close the stream, the close still hangs. Nathan Hickey was able to find a way to allow the "Client Complete Streaming Call.vi" to complete even if there are outstanding data items to read...he will post a commit to add a flag so users can choose if they want the current behavior or a forced closing of the stream (internally it will call gRPC Cancel instead of Finish). The new behavior will allow the client stream to close even if there are outstanding data items to read and any subsequents read/writes on that stream after the close will return an error and not hang.

NIthan314:
### Justification
fixes #116 
Currently, **[Client Complete Streaming Call.vi](https://github.com/ni/grpc-labview/compare/master...NIthan314:users/nhickey/add-cancel-api?expand=1#diff-673a2a4411525f55d07e3c5d4f16db4e20d3b6eda401bfe5f1748eb3a8de3cb5)** ends up calling [Finish()](https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_client_streaming_interface.html#a567f213cc0a5df2a03becda3e5711e25), which can hang if there are outstanding messages.

This change gives users the option to call [TryCancel()](https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_context.html#abd0f6715c30287b75288015eee628984) from the client instead, which doesn't hang if there are outstanding messages.

### Implementation

- Added entry point in grpc_client.cc to expose the Cancel API
- Updated Client Complete Streaming Call.vi to accept a boolean, that if true calls the Cancel API instead of finish


shivaprasad-basavaraj:
It might be better to have an explicit LV API as well to cancel an ongoing call instead of updating the 'Client Complete Streaming Call.vi' to take an additional Cancel parameter. It makes things more explicit that you are just canceling the ongoing call altogether and we can use it for Unary calls as well.

NIthan314:
@shivaprasad-basavaraj, I made an explicit LV API for Cancel and reverted 'Client Complete Streaming Call.vi'. Let me know if that works for you.

shivaprasad-basavaraj:
> @shivaprasad-basavaraj, I made an explicit LV API for Cancel and reverted 'Client Complete Streaming Call.vi'. Let me know if that works for you.

@NIthan314 The changes look good to me.

shivaprasad-basavaraj (approved):

PR #120, opened 2022-08-24
Token manager

chadallee:
I added a token manager class to manage the various grpc pointers that we're passing back and forth between LabVIEW and the DLL.

```
    /// This class serves two purposes:
    /// 1. Ensure the validity of pointers -- This class knows which tokens are valid, so if
    ///    LabVIEW passes us an invalid/stale token, we know it's invalid without ever trying to cast it.
    /// 2. Gracefully clean up pointers only when they're no longer being used by utilizing RAII (std::shared_ptr).
    ///    Client code never has to explicitly call `delete` on pointers. std::shared_ptr automatically calls
    ///    delete when there are no remaining instances of std::shared_ptr for a given pointer. Usually, the
    ///    call to DestroyTokenForPtr() will remove the final instance of std::shared_ptr from the map, and that will
    ///    cause `delete` to be called on that pointer. But if some other method in some other thread still has the std::shared_ptr
    ///    from GetPtrFromToken(), then `delete` won't be called until that other std::shared_ptr goes out of scope.
```

I used a uintptr_t as the token type so that the existing DLL entrypoints (used by the LabVIEW CLNs) will continue to work exactly the same for 32-bit and 64-bit LV.

This change extends on top of Nathan's Cancel() PR: https://github.com/ni/grpc-labview/pull/118
If his change goes in first, I think I branched correctly so that this will merge nicely.
Or we can just merge this one PR that includes his changes too.


src/token_manager.hpp
```diff
@@ -0,0 +1,64 @@
...
+    {
+    }
+
+    template <typename T>
+    template <typename TDerivedType>
+    std::shared_ptr<TDerivedType> TokenManager<T>::GetPtrForToken(TokenManager::token_type token, int32_t* status)
+    {
+        std::lock_guard<std::mutex> lock(_mutex);
+        auto ptr = _pointerForToken.find(token);
+        if (ptr == _pointerForToken.end())
```
chadallee, line 19: RFC: I don't know the right way to handle errors. There was a lot of "return -1" already in grpc_client.cc.

1. If there's a better error range to use, I'd be happy to do so. Maybe one of the [gRPC status codes](https://grpc.github.io/grpc/core/md_doc_statuscodes.html), possibly along with the `-(1000 + status)` that grpc_client already uses in some places.
2. Maybe we don't need a status here. Just return the default pointer like I already am and let the client code check for an invalid pointer. This is essentially what I've already got implemented since all of the calling code that I changed is passing `nullptr` for the `status` pointer.
3. I don't know if writing to std::cerr is a good thing here either. I don't like hardcoding non-localized strings in code.

src/token_manager.hpp
```diff
@@ -0,0 +1,64 @@
+#include "token_manager.h"
```
ccifra, line 1: should stick with .h for the file extension.  We have not been using .hpp

src/lv_interop.h
```diff
@@ -40,10 +41,9 @@ namespace grpc_labview
         }
 
         bool IsValid() { return true; }
```
ccifra, line 43: The intent of the IsValid on gRPC ID was to eventually have a managed list of all pointers that we valid.  The token_manager seems like the right thing to store all of the valid pointers.  IsValue can get the manager and return false if the pointer is not registered.  All of the entrypoints should be using CastTo and handling the null case as an error returned to LabVIEW.  I would put the use of the token manager here.

src/token_manager.hpp
ccifra: oops, it looks like you added the same file twice with two extensions.

ccifra (changes requested):

chadallee:
submit "pending" PR comments?

src/lv_interop.h
ccifra: It is safe to call IsValid.  It is not a virtual function and C++ does not care if the this pointer is valid or not.  So you can call IsValid and that can call into the token manager to ensure the pointer is still valid (not deleted, not null).

src/pointer_manager.h
```diff
@@ -0,0 +1,115 @@
...
+    template <typename T>
+    class PointerManager
+    {
+    public:
+        PointerManager<T>();
+
+        /// Register a pointer. This creates a std::shared_ptr for that pointer and stuffs it into a std::set.
+        /// The pointer will be alive until that shared_ptr and all its copies (created via GetPointer) are out of scope.
+        /// Call UnregisterPointer to allow the pointer to be destroyed.
+        T* RegisterPointer(T* ptr);
```
ccifra, line 33: Since we are using shared_ptr semantics for lifetime management, should we take in a shared pointer for registration?

I can go either way on this.  I realize that given how we are using things with CLN we are passing pointers around and Unregister makes more sense to take a pointer.

ccifra (approved):

src/pointer_manager.h
ccifra: eh, I am just going to submit as it. We can think about it some more and do a follow up if we think it is warranted.

chadallee: I think I like it just taking a raw pointer, but I could go either way as well.

I also thought about not taking a pointer at all and just telling the PointerManager to create the object.

i.e.
`grpc_labview::gPointerManager.CreateAndRegisterPointer<grpc_labview::ClientCall>(timeoutMs)`
instead of
`*callId = grpc_labview::gPointerManager.RegisterPointer(new grpc_labview::ClientCall(timeoutMs));`).

Maybe even make all the constructors for derived gRPCid classes protected so that pointers can only be created by that factory method. But that seemed like overkill.

Like you said, since we're passing naked pointers to and from the CLNs, I vote we just expose the naked pointers.

Issue #119, opened 2022-08-24
Packing and unpacking data in 64 bit grpc-labview sometimes loses data


shivaprasad-basavaraj:
Packing and unpacking data is sometimes lossy. In the attached example, we are packing and unpacking data and some data is lost as seen in the image below. This happens only with the 64-bit version.

![image](https://user-images.githubusercontent.com/35128660/186388734-3c6fed6e-f417-4318-9b35-24a4963eed23.png)

*Steps to reproduce*
- Extract the attached 'proto parsing.zip' and open 'proto debug.lvproj'
- Open the SimplestPackUnpack.vi and run it
- Observe that we lose data as shown in the image

_The project is saved in LV 2021_

[proto parsing.zip](https://github.com/ni/grpc-labview/files/9414990/proto.parsing.zip)
.

gregr-ni:
We are computing the wrong size for the first 2 clusters in UpdateMetadataClusterLayout. We're padding them to 8 bytes which would be correct if they contained any 8 byte values but they don't. Clusters are only padded to the max size of their elements. Specifically the [AlignClusterOffset ](https://github.com/ni/grpc-labview/blob/f402ef2dbbdad8321cbe0bd0079934782063fabe/src/message_element_metadata_owner.cc#L153)used to assign the clusterSize is wrong. It can't use a fixed type (currently StringValue). It should use the type of the largest element and false for "repeated".

This also means that we are copying 8 bytes beyond the end of the cluster into the message when packing and overwriting 8 bytes beyond the cluster when unpacking.

shivaprasad-basavaraj:
Fixed in release 0.4.8.4.

PR #122, opened 2022-08-25
Fix byte alignment issues with 64 bit grpc-labview

shivaprasad-basavaraj:
Fix for Issue #119 

- Updating calculation of cluster byte alignment for 64 bit grpc-labview to be based on the biggest byte alignment requirement we have seen among all the elements of the cluster. This is a recursive calculation in the case of nested clusters.

Tested the fix with the following configurations for the payload we are packing/unpacking.
- Nested clusters (#119)
- Clusters with strings
- Clusters with arrays
- Clusters with arrays of clusters
- Clusters with arrays of strings

### DIFF
--- a/src/cluster_copier.cc
+++ b/src/cluster_copier.cc
@@ -223,7 +223,7 @@ namespace grpc_labview {
                 int x = 0;
                 for (auto str : repeatedNested->_value)
                 {
-                    auto lvCluster = (*array)->bytes<LVCluster*>(x * clusterSize);
+                    auto lvCluster = (LVCluster**)(*array)->bytes(x * clusterSize, nestedMetadata->alignmentRequirement);
                     *lvCluster = nullptr;
                     CopyToCluster(*str, (int8_t*)lvCluster);
                     x += 1;
@@ -667,7 +667,7 @@ namespace grpc_labview {
 
                     for (int x = 0; x < count; ++x)
                     {
-                        auto data = (*array)->bytes<LVCluster*>(nestedMetadata->clusterSize * x);
+                        auto data = (LVCluster**)(*array)->bytes(nestedMetadata->clusterSize * x, nestedMetadata->alignmentRequirement);
                         auto nested = std::make_shared<LVMessage>(nestedMetadata);
                         repeatedValue->_value.push_back(nested);
                         CopyFromCluster(*nested, (int8_t*)data);
--- a/src/cluster_copier.cc
+++ b/src/cluster_copier.cc
@@ -667,7 +667,7 @@ namespace grpc_labview {
 
                     for (int x = 0; x < count; ++x)
                     {
-                        auto data = (LVCluster**)(*array)->bytes(nestedMetadata->clusterSize * x, nestedMetadata->alignmentRequirement);
+                        auto data = (LVCluster*)(*array)->bytes(nestedMetadata->clusterSize * x, nestedMetadata->alignmentRequirement);
                         auto nested = std::make_shared<LVMessage>(nestedMetadata);
                         repeatedValue->_value.push_back(nested);
                         CopyFromCluster(*nested, (int8_t*)data);
--- a/src/lv_interop.h
+++ b/src/lv_interop.h
@@ -46,6 +46,8 @@ namespace grpc_labview
         virtual ~gRPCid() { }
     };
 
+    int AlignClusterOffset(int clusterOffset, int alignmentRequirement);
+
     //---------------------------------------------------------------------
     // LabVIEW definitions
     //---------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


src/message_element_metadata_owner.cc
```diff
@@ -120,37 +125,58 @@ namespace grpc_labview {
...
-                clusterOffset += ClusterElementSize(element->type, element->isRepeated);
+                clusterOffset = AlignClusterOffset(clusterOffset, element->type, element->isRepeated);
+                element->clusterOffset = clusterOffset;
+                int alignmentRequirement = 0;
+                auto elementSize = ClusterElementSize(element->type, element->isRepeated);
+                clusterOffset += elementSize;
+                if (element->isRepeated)
+                {
+                    // This should be the alignmentRequirement of the LV1DArray struct which should be 4 (contains int32 and int8)
+                    alignmentRequirement = 4;
```
shivaprasad-basavaraj, line 166: We might have to do something similar for Strings and BytesValue as well since they will also be struct of an int32 count and an int8 single-element array. So the size of the struct will be 8 but byte alignment should be 4.

shivaprasad-basavaraj: I had missed the fact that for strings/bytearrays we would have a pointer to the LStr here and not the actual struct itself. So this would be 8 on 64-bit. I have fixed this assumption for arrays as well.

src/lv_interop.h
```diff
@@ -93,6 +93,20 @@ namespace grpc_labview
             return (T*)(rawBytes + byteOffset);
     #endif
         }
+
+        void* bytes(int byteOffset, int byteAlignment)
```
shivaprasad-basavaraj, line 97: The bytes methods above made assumptions about where the raw data of the array would start.  This assumption broke for an array of cluster since we were trying to read them as a pointer to an LVCluster which meant it was 8 bytes in size for 64 bit but that the assumption that the rawBytes would always be 8 byte aligned wouldn't hold good in that case. We should probably not make assumptions at all about byte alignment based on type and always have the user pass in the byte alignment requirements for the elements of the array.

I kept the above methods for now since they are used in a lot of places, We will fix them in a separate PR.

src/cluster_copier.cc
```diff
@@ -667,7 +667,7 @@ namespace grpc_labview {
 
                     for (int x = 0; x < count; ++x)
                     {
-                        auto data = (*array)->bytes<LVCluster*>(nestedMetadata->clusterSize * x);
+                        auto data = (LVCluster**)(*array)->bytes(nestedMetadata->clusterSize * x, nestedMetadata->alignmentRequirement);
```
gregr-ni, line 670: This is not `LVCluster**`. It is just `LVCluster*`. It doesn't change execution since all you do with it is cast it to `int8_t*` below, but it is wrong.

src/lv_interop.h
gregr-ni: The original code would have been more appropriate as `<LVCluster>` but that still doesn't work because a clusters alignment requirement can't be determined just from its size.  The templatized `bytes` is fine for any other type. You might be able to use static asserts or template specialization to make sure it LVCluster is not allowed.

src/lv_interop.h
```diff
@@ -93,6 +93,20 @@ namespace grpc_labview
...
+        {
+#ifndef _PS_4
+            int reminder = 4 % byteAlignment;
+            if (reminder == 0)
+            {
+                return (void*)(rawBytes + byteOffset);
+            }
+            return (void*)(rawBytes + byteAlignment - reminder + byteOffset); // 8-byte aligned data
```
gregr-ni, line 105: We've got at least 4 copies of alignment calculation that are not all the same. The templatized `bytes` functions are simpler because they know arrays are always maximally aligned and start with a 4 byte size. If you are writing array specific code, you can continue this. The code doesn't need to do a modulo. It can look like the previous function. Then to avoid duplication the templatized functions can call the new function passing in `sizeof` as `alignmentRequirement`.

If you take it one step further, all 3 can call `AlignClusterOffset` and remove even more duplication.

src/message_element_metadata_owner.cc
```diff
@@ -69,6 +69,16 @@ namespace grpc_labview {
         return 0;
     }
 
+    int AlignClusterOffset(int clusterOffset, int alignmentRequirement)
```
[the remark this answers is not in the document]
gregr-ni: This code is still wrong for 32-bit.

gregr-ni (changes requested):

src/message_element_metadata_owner.cc
```diff
@@ -120,37 +125,58 @@ namespace grpc_labview {
...
                 UpdateMetadataClusterLayout(nestedMetadata);
+                int alignmentRequirement = 0;
+                int elementSize = 0;
                 if (element->isRepeated)
-                {                
-                    clusterOffset += ClusterElementSize(element->type, element->isRepeated);
+                {
+                    elementSize = ClusterElementSize(element->type, element->isRepeated);
+                    // This should be the alignmentRequirement of the LV1DArray struct which should be 4 (contains int32 and int8)
+                    alignmentRequirement = 4;
```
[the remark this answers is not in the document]
shivaprasad-basavaraj: I have updated the ClusterElementSize() method to use the sizeof void pointer where it would be pointer types underneath.

src/lv_interop.h
shivaprasad-basavaraj: I have updated the bytes() template methods to have static assert when used with struct/class types which would require the explicit calculation of byteAlignment. All of them now use the AlignClusterOffset method to calculate byte alignment. 

There was one other usage of the bytes() template method which passed a struct in grpc_interop.cc. I have updated that method as well to pass in an explicit byteAlignment instead.

src/message_element_metadata_owner.cc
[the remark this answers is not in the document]
shivaprasad-basavaraj: I have moved the `_PS_4` check to the other `AlignClusterOffset` which actually takes in a byteAlignment. 

The _PS_4 flag is set in the cmake file [here](https://github.com/ni/grpc-labview/blob/d44834c2567500d837f826fa029a828dad1fa58e/CMakeLists.txt#L22).

gregr-ni (approved):

ccifra (approved):

src/message_element_metadata_owner.cc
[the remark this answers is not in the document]
shivaprasad-basavaraj: This change was in the preliminary PR where I had misunderstood how the LV1DArray would be stored in memory. I have since updated the changes so that we use the correct element size and alignment requirement.

src/message_element_metadata_owner.cc
```diff
@@ -120,37 +125,58 @@ namespace grpc_labview {
...
                 auto nestedMetadata = FindMetadata(element->embeddedMessageName);
                 UpdateMetadataClusterLayout(nestedMetadata);
+                int alignmentRequirement = 0;
+                int elementSize = 0;
                 if (element->isRepeated)
-                {                
-                    clusterOffset += ClusterElementSize(element->type, element->isRepeated);
+                {
+                    elementSize = ClusterElementSize(element->type, element->isRepeated);
+                    // This should be the alignmentRequirement of the LV1DArray struct which should be 4 (contains int32 and int8)
```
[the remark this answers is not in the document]
shivaprasad-basavaraj: This change was in the preliminary PR where I had misunderstood how the LV1DArray would be stored in memory. I have since updated the changes so that we use the correct element size and alignment requirement.

PR #125, opened 2022-08-29
Added RT build script and updated other scripts to upload artifacts

abahetik:
Files Added:
1. Build_on_rt.yml : This builds rt so using and an rt compiler. This script will also upload its artifact.
2. CI.yml: This script will be used to trigger all other build workflows.
3. nilrt-x86_64.cmake: This cmake file is used to set variables that are needed by the Linux RT Toolchain.

Updated ymls:
1. windows_x64_build: Added functionality to run on Pull Requests and it also uploads its artifacts.
2. windows_x84_build: Added functionality to run on Pull Requests and it also uploads its artifacts.
4. build_on_linux: Added functionality to run on Pull Requests and it also uploads its artifacts.


ni-sujain (approved):
looks good

PR #131, opened 2022-09-01
Set up CI with Azure Pipelines

abahetik:


abahetik:
This PR was a Test to see how do we connect to github actions through Azure pipline

PR #138, opened 2022-09-08
LabVIEW was crashing if the stream had already been closed.

NIbbuchanan:
This fix checks if id is valid and if not, it signals the occurrence (to avoid LV hanging and returns immediately so we don't try using a null reference.


src/grpc_client.cc
```diff
@@ -441,6 +441,12 @@ LIBRARY_EXPORT int32_t ClientBeginReadFromStream(grpc_labview::gRPCid* callId, g
     auto reader = callId->CastTo<grpc_labview::StreamReader>();
     auto call = callId->CastTo<grpc_labview::ClientCall>();
     auto occurrence = *occurrencePtr;
+	
+	if (!reader || !call)
```
ni-sujain, line 445: just fix the indentation here

ni-sujain:
Can you attach an example for this. We can add this as a test case

ni-sujain (approved):

src/grpc_client.cc
NIbbuchanan: Fixed. I also removed an extra call to gPointerManager.RegisterPointer(client); on line 171 since we already register the pointer two lines later since this was causing a leaked reference that wasn't cleaned up correctly. @chadallee is also investigating a related LV crash when we close the VI (ie. the DLL goes out of scope and destructors are called).

NIbbuchanan: Here is a sample to reproduce the crash. With the fix, it no longer crashes (until you unload the DLL by closing the VI). As mentioned earlier @chadallee is investigating this issue.
[test_grpcServerWithFlexibleCloseOptions.zip](https://github.com/ni/grpc-labview/files/9548800/test_grpcServerWithFlexibleCloseOptions.zip)

Issue #140, opened 2022-09-12
LabVIEW crashes on close after running grpc test VI


NIbbuchanan:
With the latest updates from 0.4.9.1 (and the fix for the crash mentioned here: https://github.com/ni/grpc-labview/pull/138), my LabVIEW crashes when I run the attached example and then close the VI. It should go to the LabVIEW splash screen, but instead the LV process disappears without a crash dialog or anything. This is causing our test suite to fail everything that uses grpc. A co-worker with the same version of LV and INI file doesn't have this problem, but it happens on my dev machine as well as on all our test machines. I can provide an IP address to developers that want to remote into a machine that reproduces this issue if needed.
[test_grpcCrashLVAfterCloseTest.zip](https://github.com/ni/grpc-labview/files/9551185/test_grpcCrashLVAfterCloseTest.zip)

NIbbuchanan:
We did some initial investigation and it appears there may be multiple deallocations of the same pointer. @chadallee found that a "timer_thread" may be de-allocating memory on close, but we haven't figured out where the problem is yet.

chadallee:
The crash occurs because a `timer_thread` is still running after the test VI is finished running.
Moreover, it's still running when the test VI is closed, which also unloads the labview_grpc_server.dll.  Once the DLL has been unloaded, that thread is still running and it wakes up and crashes.

Here is the callstack of the thread that's still running even when the VI is idle.

```
 # Child-SP          RetAddr           Call Site
00 0000008c`c29ff710 00007fff`8c27e1c0 labview_grpc_server!absl::lts_20211102::synchronization_internal::Waiter::Wait+0xba [C:\dev\github-grpc-labview\third_party\grpc\third_party\abseil-cpp\absl\synchronization\internal\waiter.cc @ 385] 
01 0000008c`c29ff790 00007fff`8c27d098 labview_grpc_server!AbslInternalPerThreadSemWait_lts_20211102+0xd0 [C:\dev\github-grpc-labview\third_party\grpc\third_party\abseil-cpp\absl\synchronization\internal\per_thread_sem.cc @ 92] 
02 0000008c`c29ff7e0 00007fff`8c27349a labview_grpc_server!absl::lts_20211102::synchronization_internal::PerThreadSem::Wait+0x28 [C:\dev\github-grpc-labview\third_party\grpc\third_party\abseil-cpp\absl\synchronization\internal\per_thread_sem.h @ 113] 
[...]

PR #143, opened 2022-09-15
Ensure gRPC server dll is never unloaded from the process once it is loaded

shivaprasad-basavaraj:
Fix for Issue #140 

There is a potential crash with labview-grpc when the grpc dll is unloaded if the user hasn't shut down everything correctly when running his project which makes use of labview-grpc. This would happen because some of the threads created by the core gRPC layer might still be around and when the gRPC dll is unloaded from underneath these threads, the application would crash. There are more discussions about this issue in #140. This is a [known issue](https://github.com/grpc/grpc/issues/21902#issuecomment-630285928) with gRPC C++.

This PR makes changes so that the gRPC dll loads itself once so that it won't be unloaded till process exit.


ni-sujain (approved):
I hope it does not create a problem for RT cases. Also should we consider Freelibrary() as well at some point or it will get handle by process kill?

shivaprasad-basavaraj:
> I hope it does not create a problem for RT cases. Also should we consider Freelibrary() as well at some point or it will get handle by process kill?

I tested on RT and it worked without any issues. With this change the intention is that the gRPC DLL is never unloaded once loaded. So we shouldn't call Free library. It will happen automatically on process exit.

PR #155, opened 2022-09-28
Updated ci.yml workflow filter

abahetik:
Changed the ci.yml to:

1. Run only on the latest commit for pull request as well as master branch
2. Disabled release-artifact.yml to run on forked branches

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -8,6 +8,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
ni-sujain, line 12: does it mean that if we have build wrt. one commit on master still in progress and we do a second commit, it will cancel the ongoing build and start a new build

ni-sujain (approved):

.github/workflows/ci.yml
abahetik: Yes.

PR #160, opened 2022-10-04
Use sync object to protect call list within client

ericgross2:
This is a simplistic fix for the bug #161  we are seeing where we crash when sending multiple gRPC commands at one time. The issue appeared to be missing synchronization protecting the std::list containing the active calls on a client.

Note: A cleaner fix would be to encapsulate adding/removing to the list as a method that internally grabs the sync object.


ericgross2:
This fixes #161

gregr-ni (approved):

Issue #144, opened 2022-09-15
Specifying CLFN paths on diagram reduces usability


gregr-ni:
When Call Library Function nodes use a path from the diagram, LV builds can't know the location of the library and automatically include it in the built output. This is mitigated in the MSDK service case by explicit post build code in the template projects, but it will be a constant source of issue for general gRPC usage.

Currently the libraries ship in the locations below (relative to the LV directory) and `Get Server DLL Path.vi` produces the correct path for each platform.

- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\Linux\liblabview_grpc_server.so
- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\LinuxRT\liblabview_grpc_server.so
- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\Win32\labview_grpc_server.dll
- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\Win64\labview_grpc_server.dll

Instead the libraries should ship in the locations below, `Get Server DLL Path.vi` should be deleted, and all CLFNs should be configured to reference the equivalent of `<vilib>:\gRPC\LabVIEW gRPC Library\Libraries\labview_grpc_server*.*`.

- Targets\NI\RT\Linux\vi.lib\gRPC\LabVIEW gRPC Library\Libraries\labview_grpc_server64.so
- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\labview_grpc_server32.dll
- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\labview_grpc_server64.dll
- vi.lib\gRPC\LabVIEW gRPC Library\Libraries\labview_grpc_server64.so

> Note: When first configuring the CLFN with the wildcards, the VI may claim the function can't be found. However, if you save and reload the VI in this state, it fixes itself. This appears to be a bug in LV. I observed it in LV 2021, not sure about other versions.

NIbbuchanan:
[...]

Issue #75, opened 2022-03-27
Make Get Server DLL Path.vi PPL compatible


sr33-multics:
When built in to PPLs the [Get Server DLL Path.vi](https://github.com/ni/grpc-labview/blob/master/labview%20source/gRPC%20lv%20Support/Server%20API/Server/Get%20Server%20DLL%20Path.vi) does not resolve the dll path correctly.
A case structure similar to this would help resolve it.
![ppl](https://user-images.githubusercontent.com/29126754/160279669-106faa17-5a82-4dbc-ab08-70c1cc89be4c.png)

NIbbuchanan:
I would vote for not using custom lookup paths and use the default search location LV already has for finding DLLs and shared libraries so the Call Library Function can just use labview_grpc_server.*. This makes the install a little trickier than just putting all the DLLs in a Libraries folder. On windows, it should go under LabVIEW/resource folder (use the correct 32/64-bit depending on which LV they install to. We could keep them in the libraries folder so it's easy for the user to find them and copy to other targets as needed, but CLF should not take a path and should instead use the default LV search path...like all our other driver DLLs do.

pakdev:
Copied your idea into #164, which is now merged - thanks! This issue can now be closed

shivaprasad-basavaraj:
The fix is part of the [0.5.1.2 release](https://github.com/ni/grpc-labview/releases/tag/v0.5.1.2).

PR #164, opened 2022-10-05
Modify dll location if running in PPL

pakdev:
Implements changes described in issue #75. While not the ideal solution being worked out in #144, it at least unblocks us from using this library in Record/Replay


shivaprasad-basavaraj:
@pakdev Looks like the `Get Server DLL Path.vi` has been saved in LabVIEW 2020. All our VIs should be saved in LabVIEW 2019 since that's what we use to build the VIPM packages. The build_release_artifacts stage actually failed because of this but isn't being reported correctly. I have filed an issue to fix this. (#169) 

Also looks like there are issues with how the VI has been edited. It added a dependency on a `gprc-lvsupport-release.lvlib` from `Get Server DLL Path.vi` as shown in the image below. The grpc-lvsupport.lvlib gets built into the lvsupprot-release.lvlib and as such shouldn't have any dependency on the release lvlib.

image

shivaprasad-basavaraj (changes requested):
Added a comment about issues with the changes to the vi.

pakdev:
> @pakdev Looks like the `Get Server DLL Path.vi` has been saved in LabVIEW 2020. All our VIs should be saved in LabVIEW 2019 since that's what we use to build the VIPM packages. The build_release_artifacts stage actually failed because of this but isn't being reported correctly. I have filed an issue to fix this. (#169)
> 
> Also looks like there are issues with how the VI has been edited. It added a dependency on a `gprc-lvsupport-release.lvlib` from `Get Server DLL Path.vi` as shown in the image below. The grpc-lvsupport.lvlib gets built into the lvsupprot-release.lvlib and as such shouldn't have any dependency on the release lvlib.
> 
> image

Whoops, should be fixed now. Thanks!

shivaprasad-basavaraj (approved):
There was another issue with the `Get Server DLL Path.vi` where the reentrancy mode was set to non-reentrant. I went ahead and updated it to Preallocated clone like it was before. Also, the lvproj needed to be saved with the new dependency to check if the VI was in a PPL.

PR #167, opened 2022-10-07
Proto Parser Improvements

### LINKED ISSUE #73: Importing other proto definitions inside proto file
Hello there,

It seems to me that the import keyword is not supported yet for the proto file parsing. Is this something which is already considered to be done or I just have to create one big proto file from the linked proto files? I was trying to import the proto file for Opentelemetry and as it is separated into multiple proto file so it did not work. 

Thank you in advance!

-Imre

ccifra:
Add support for include parths for imported .proto files
Surface parse errors to the proto parser VI

Fixes #73


ccifra:
You can now specify any number of search paths that will be used to find imported .proto files.
Main UI now looks like this:
image

ccifra:
If there are parse errors, the file name, line number and error will be reported in the error cluster of the parser.

src/proto_parser.cc
```diff
@@ -164,6 +178,33 @@ LIBRARY_EXPORT int LVGetgRPCAPIVersion(int* version)
...
+    *parser = new grpc_labview::LVProtoParser();
+    return 0;
+}
+
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+LIBRARY_EXPORT int LVAddParserSearchPath(grpc_labview::LVProtoParser* parser, const char* searchPath)
+{
+    parser->AddSearchPath(searchPath);
```
shivaprasad-basavaraj, line 196: We could add a nullptr check here similar to the other exports here. Same for the function below.

src/proto_parser.cc
```diff
@@ -164,6 +178,33 @@ LIBRARY_EXPORT int LVGetgRPCAPIVersion(int* version)
...
+//---------------------------------------------------------------------
+LIBRARY_EXPORT int LVImportProto2(grpc_labview::LVProtoParser* parser, const char* filePath, const char* searchPath)
+{
+    parser->Import(filePath, searchPath);
+    return 0;    
+}
+
 //---------------------------------------------------------------------
 //---------------------------------------------------------------------
 LIBRARY_EXPORT int LVImportProto(const char* filePath, const char* searchPath, grpc_labview::LVProtoParser** parser)
```
shivaprasad-basavaraj, line 210: Would this function still be used anywhere? If not we could probably just replace this with the new implementation instead of creating a new entry point.

shivaprasad-basavaraj (approved):

PR #172, opened 2022-10-11
Client APIs name improvements

ccifra:
Chang the naming of the generated client APIs to be "SeviceName MethodName" so it is easier for clients to use.
The containing folder names in the project are also updated.
Naming for streaming services has also been improved as follows:
* "Begin ServiceName MethodName"
* "End ServiceName MethodName"
* "ReadStream ServiceName MethodName"
* "WriteStream ServiceName Method"

I am not too happy with the ReadStream and WriteStream names but didn't have a name that I thought was much better.  I am willing to change those before this PR goes in.  I would very much like these names to be settled.


labview source/Client Server Support New/gRPC Scripting Tools/gRPC Template Creation Utility.lvlib
```diff
@@ -269,6 +270,9 @@
 		<Item Name="Translate gRPC Method FullName.vi" Type="VI" URL="../Proto Parser API/LabVIEW API/Translate gRPC Method FullName.vi"/>
 		<Item Name="Get Message TypeUrl.vi" Type="VI" URL="../Proto Parser API/LabVIEW API/Get Message TypeUrl.vi"/>
 		<Item Name="Group Services.vi" Type="VI" URL="../Proto Parser API/LabVIEW API/Group Services.vi"/>
+		<Item Name="Create Proto Parser.vi" Type="VI" URL="../Proto Parser API/LabVIEW API/Create Proto Parser.vi"/>
```
ni-sujain, line 273: are these new helper vi's which are being added to the project, I dont see them in files added here in the PR

ni-sujain:
We can consider 
"ReadStream ServiceName MethodName" -> "ResponseStreamWrite SerivceName MethodName"
"WriteStream ServiceName MethodName" -> "RequestStreamRead SerivceName MethodName"

If that makes it better, otherwise its ok.

shivaprasad-basavaraj (approved):

Issue #176, opened 2022-10-22
LabVIEW Crash when the gRPC Client calls a service not exist in the server


NeilChien:
The situation mentioned just like this issue subject said, LabVIEW directly crashes if the service from the gRPC client does not exist at the gRPC server.
I built a simple example to reproduce this situation, please run "Crash Test.vi" to see this case. In this simple example, I run Server service from gRPC_Example_1 and use the Client of gRPC_Example_2 to communicate with the service.

[gRPC Crash LabVIEW.zip](https://github.com/ni/grpc-labview/files/9843859/gRPC.Crash.LabVIEW.zip)

I think at least we should protect LabVIEW to be crashed, then have some ability to let the designer could protect Server service communication come from unexpected clients or service-request.

Installed Version : 
LabVIEW : 2021 SP1
gRPC-LabVIEW : 0.5.0.2

NeilChien:
@gaoruhao FYI

shivaprasad-basavaraj:
Fixed in 0.5.1.3 release.

PR #179, opened 2022-10-25
Fix for crash when server doesn't implement client method

shivaprasad-basavaraj:
Fix for Issue #176 

**Issue**
With the issue described in the bug LabVIEW would crash if the gRPC server doesn't implement the method that the client is calling. Debugging the issue I found that the crash happened [here](https://github.com/ni/grpc-labview/blob/b94680b19034b4e6edae7f9746bd967f37c1c329/src/event_data.cc#L198) with an error saying there were too many operations in progress. This seemed to be because we had already done a Read on the stream and we expected a write before we called Finish() on the stream.

**Fix**
Moving the check to the Read case above fixed the issue so that if we run across a method that isn't implemented we call finish on the stream and set the error code without ever trying to use the stream.

**Testing**
Tested the fix against different types of calls (Unary and different streaming types).

### DIFF
--- a/src/event_data.cc
+++ b/src/event_data.cc
@@ -204,8 +204,9 @@ namespace grpc_labview
             }
             else
             {
+                _status = CallStatus::Finish;
                 _stream.Finish(grpc::Status(grpc::StatusCode::UNIMPLEMENTED, ""), this);
-            }       
+            }
         }
         else if (_status == CallStatus::Writing)
         {
--- a/src/event_data.cc
+++ b/src/event_data.cc
@@ -172,8 +172,17 @@ namespace grpc_labview
             // part of its FINISH state.
             new CallData(_server, _service, _cq);
 
-            _stream.Read(&_rb, this);
-            _status = CallStatus::Process;
+            auto name = _ctx.method();
+            if (_server->HasRegisteredServerMethod(name) || _server->HasGenericMethodEvent())
+            {
+                _stream.Read(&_rb, this);
+                _status = CallStatus::Process;
+            }
+            else
+            {
+                _status = CallStatus::Finish;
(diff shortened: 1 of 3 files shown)


src/event_data.cc
```diff
@@ -172,8 +172,17 @@ namespace grpc_labview
...
+            auto name = _ctx.method();
+            if (_server->HasRegisteredServerMethod(name) || _server->HasGenericMethodEvent())
+            {
+                _stream.Read(&_rb, this);
+                _status = CallStatus::Process;
+            }
+            else
+            {
+                _status = CallStatus::Finish;
+                _stream.Finish(grpc::Status(grpc::StatusCode::UNIMPLEMENTED, ""), this);
```
ni-sujain, line 184: Q: Do we advertise the grpc::status somehow? when we say UNIMPLEMENTED will user see that?

if we are showing, then we may want to pass some error message like grpc::Status(grpc::StatusCode::UNIMPLEMENTED, "Method not implemented yet");

src/event_data.cc
```diff
@@ -195,7 +204,7 @@ namespace grpc_labview
             }
             else
             {
-                _stream.Finish(grpc::Status::CANCELLED, this);
+                _stream.Finish(grpc::Status(grpc::StatusCode::UNIMPLEMENTED, ""), this);
```
ni-sujain, line 207: do we also want to set _status = CallStatus:Finish here

ni-sujain (approved):
looks good

src/event_data.cc
shivaprasad-basavaraj: This error status will make its way back to a LV error cluster and you will see the error below.
image

src/event_data.cc
shivaprasad-basavaraj: Good catch. We should never get into this scenario since we have already had a check during the read phase itself but set the status anyway.

PR #180, opened 2022-10-26
Fix a hang in SetResponseData() and a crash in CompleteClientUnaryCall2()

chadallee:
Primary issue:
I saw an occasional hang in `SetResponseData()` -> `CallData::Write()` when running some of our test suites.
That `CallData::Write()` method was stuck waiting forever on the `_writeSemaphore()`. When we cancel calls, the `CallData::Proceed()` method gets invoked with `ok=false`. If that happens while we're in the middle of a write (i.e. `_status == CallStatus::Writing`), then that `::Proceed()` method skips the `Writing` state and goes straight to PendingFinished. Skipping the `Writing` state means we never call `_writeSemaphore.notify()`, which means the active `SetResponseData()` call to `CallData::Write()` never finishes.

Secondary issue:
While trying to write a simpler reproduction VI for that first issue (our test suite takes ~25 minutes to run and this hang was only reproducing like one out of 5-10 times), I managed to crash LabVIEW pretty frequently. The problem was that I called `DestroyClient` in parallel with a call to `Client Unary Call.vim`. `Client Unary Call.vim` calls into `ClientUnaryCall()` and `CompleteClientUnaryCall2()`. Specifically, I was seeing the crash at the end of `CompleteClientUnaryCall2()` where we call `call->_client->ActiveClientCalls.remove(call.get());`. When it crashed, I found that the `LabVIEWgRPCClient` object stored in `call->_client` had been deleted. So instead of holding onto a raw pointer, I made ClientCall hold a std::shared_ptr so that it can't get deleted out from under us.


chadallee:
I don't think I have write access to this project. What else do I need to do to get this merged?

PR #196, opened 2022-11-22
Added the TestTemplate and new Test VIs

abahetik:
This PR added:
1. Test_TemplateVI.vit : A template VI which all Server and Client VIs can use.
image
2. Test_HelloWorld.vit
image
3. Test_RouteGuide_Unary.vi 
image
4. Test_RouteGuide_BidirectionalStreaming.vi 
image
5. Test_RouteGuide_ResponseStreaming.vi 
image
6. Test_RouteGuide_ResponseStreaming.vi 
image
7. Test_ServiceNotFoundInServer.vi - This uses the folder to ensure that the Server does not crash when that service is not available.
image

This PR modified:
1. run_tests.py - Changed the way of reading lst file. Added check to test whether the testVI path is correct or wrong
2. HelloWorldTest.vi - This VI has been replaced using the new template
3. The changes in the examples Client VIs has connected the terminals(input, output and error out)
4. Tests.lst - added the new tests


tests/run_tests.py
```diff
@@ -22,7 +22,7 @@ def run_all_tests():
     test_file = os.path.join(test_directory , "Tests.lst")
     failed_test_results = ""
     with open(test_file, 'r') as file:
-        all_tests = file.readlines()
+        all_tests = file.read().splitlines()
```
abahetik, line 25: Made this change to prevent from getting **'/n'**

ni-sujain (approved):
- for Service not found test case, remove the comments blocks inside the Test VI
- also rename the CrashTest.vi
- may be put a readme on what the test is doing as it is different from the template cases we are having

shivaprasad-basavaraj:
What were the changes to the route guide example? I am guessing the client changes were wiring the error out. But what was the change to server VI? You can document that as part of the PR description.

shivaprasad-basavaraj (approved):

shivaprasad-basavaraj:
Also did you have to add the outputs on the connector pane of the route guide client VIs? I don't think we should do that since those client runners are top-level VIs which shouldn't have anything on the connector pane. Maybe just doing a VI server RunVI similar to the way we run the server is better. I already added a helper VI to get control values from a VI ref in my cancellation PR (#197)

abahetik:
Changed Test VIs to:
1. Unary Test
image
2. Response Streaming Test
image
3. Bidirectional Streaming
image
4. Request Streaming
image

The Client VIs have added error in and error out control. And the default values have been changed

PR #197, opened 2022-11-25
Support for cancelling ongoing calls

shivaprasad-basavaraj:
Added support for canceling ongoing calls from the client side. The changes included in this PR are described below:

- Added new APIs to create and close a client context
- Added a new cancellation API which calls `TryCancel()` on the ClientContext that was passed in
- Modified the existing APIs which initiate a client call to take in the client context as a parameter. If we cannot find the context that was passed in, we will create a new one. We can use this context field to implement call context propagation in the future.
- Updated the client code generation template VIs to wire in the context as an optional terminal

With this change, I had to update the connector pane pattern for `Client Unary Call.vim` since almost all the inputs were already occupied and I didn't have space to pass in the contextId. This change could potentially break clients for grpc-labview but the fix is pretty simple. They just have to go to the call sites of `Client Unary Call.vim` and do a right click -> Relink to SubVI to fix the issue. I have updated the existing examples with the fix.

**Tests**
Added new tests to check for cancellation in both Unary and Streaming calls. The tests are based on the route guide example.


ni-sujain (approved):

ni-sujain:
Once we publish the release notes, lets add that this will break the connector pane

PR #214, opened 2022-12-19
Sync GitHub Issues to Azdo

abahetik:
This PR adds a new workflow :
sync_github_issues_to_azdo: This workflow will create an equivalent AzDo bug for the GitHub issues
and
Fixed pushing the automated VIPB bumps to master


shivaprasad-basavaraj:
This PR would duplicate only new issues right? What happens to the issues that are already raised in GitHub?

abahetik:
> This PR would duplicate only new issues right? What happens to the issues that are already raised in GitHub?

@shivaprasad-basavaraj It will get triggered for all issue types i.e opened, edited, deleted, closed, reopened, labeled, unlabeled, assigned

shivaprasad-basavaraj:
> > This PR would duplicate only new issues right? What happens to the issues that are already raised in GitHub?
> 
> @shivaprasad-basavaraj It will get triggered for all issue types i.e opened, edited, deleted, closed, reopened, labeled, unlabeled, assigned

So the issues won't be duplicated in AzDO unless they are interacted with somehow right? Is that what we are going for?

abahetik:
> > > This PR would duplicate only new issues right? What happens to the issues that are already raised in GitHub?
> > 
> > 
> > @shivaprasad-basavaraj It will get triggered for all issue types i.e opened, edited, deleted, closed, reopened, labeled, unlabeled, assigned
> 
> So the issues won't be duplicated in AzDO unless they are interacted with somehow right? Is that what we are going for?

@shivaprasad-basavaraj The issue won't get duplicated even when edits are made. It will just update its equivalent AzDo item. However, I think the existing issues might need to be modified to trigger the workflow

shivaprasad-basavaraj (approved):

Issue #217, opened 2022-12-26
gRPC response not shown consistently


yijueli:
gRPC version: v0.5.2.3

This is an issue that highly effects the ADAS system stability. After long runtime or some wrong connections , we cannot get gRPC response unless the VI is closed and opened again.

It's hard to reproduce the long-run case which worries us a lot. But l made an example([gRPC Response Issue.zip](https://github.com/ni/grpc-labview/files/10312284/gRPC.Response.Issue.zip)) to reproduce the issue after a wrong connection. Try the steps below one more times if you cannot reproduce the issue for the first time. Steps:
1. Open the Test. lvproj and connect a RT target.
2. Open gRPC Client Test.vi and Run Service.vi.
![image](https://user-images.githubusercontent.com/105094769/209760853-54604a6b-579c-474e-bc12-944a83228b37.png)
3. Run Run Service.vi. 
4. In gRPC Client Test.vi, input "localhost" and the grpc port from  Run Service.vi into the address as the picture below. You can get the expected -1014 error if you set an enough long timeout.
![image](https://user-images.githubusercontent.com/105094769/209761109-b2868403-8aec-49ab-84e7-5bc877f51032.png)
5. Stop the gRPC Client Test.vi and input the right RT address and the grpc port into the address.
Then you can see the issue: **_NO ERROR, NO RESPONSE!_** You can stop the vi and rerun it, the issue still exists.
![image](https://user-images.githubusercontent.com/105094769/209761533-7db21a9f-eade-43d7-89dc-d8a33cddd484.png)
6. Close the gRPC Client Test.vi and reopen it, it works normally.

In addition, l found the other issue during the process. When l connected a wrong port which did not provide
 gRPC services and requested Greeter service in a loop, it sometimes threw -1004 error and sometimes -1014 error alternately. The problem misled us to locate other issues.
![image](https://user-images.githubusercontent.com/105094769/209494275-8737128c-2ba3-4481-a8c6-7316f8552eec.png)
![image](https://user-images.githubusercontent.com/105094769/209493764-28aa9bec-10f7-46cd-b3cf-6875870a0169.png)

AB#2262093

AB#2272017

shivaprasad-basavaraj:
@yijueli I tried to reproduce this behavior but when I run the unary call of the helloworld project in a loop with the wrong address I consistently get an error -1014 mentioning it failed to connect to the address. You can see my modification to the helloworld client below. Does this resemble the client VI you were running?

[...]

PR #223, opened 2023-01-27
Dont signal occurrence if client call not active

ni-sujain:
fix for issue #217 

The issue is that client can run into a situation where an inactive client call(e.g. due to vi aborts) will end up signaling the lv occurrence from async thread for an active client call(occurrence is getting re-used) and "wait for occurrence" unblocks without reading the response from server and we will see no response for the active call.

The fix involves clearing the active client calls list on vi abort and signal occurrence only when the current call is active. **This fix will work for most cases but it still can run into some race conditions.**

Alternatives considered:
- Somehow reset the occurrence on vi abort. Could not find a way to do this. Even if a signal the occurrence in `ClientCleanupProc` it doesnot work.
- Reset occurrence before we run ClientUnaryCall CLFN node. Tried _Wait on occurrence_  primitive with zero timeout,  _Set occurence_ primitve but they also did not work. as even if we reset, the async c++ can signal before we go inside new ClientUnaryCall CLFN.
- Do a wait() on blocking client call but this takes away some async behavior.  

testing:
I tried with the attached example from the issue. It runs fine.


src/grpc_client.cc
```diff
@@ -192,6 +193,22 @@ int32_t CloseClient(grpc_labview::LabVIEWgRPCClient* client)
     return 0;
 }
 
+// Signal a lv occurence for a client call from async c++ thread
+void SignalOccurenceForClientCall(grpc_labview::ClientCall* clientCall)
```
shivaprasad-basavaraj, line 197: Should we name this method `CheckActiveAndSignalOccurence` to signify that we will fire the occurrence only if the client call is active?

shivaprasad-basavaraj:
We should add this scenario as an auto test as well.

src/grpc_client.cc
```diff
@@ -218,6 +235,8 @@ int32_t ClientCleanUpProc(grpc_labview::gRPCid* clientId)
         {
             (*activeClientCall)->Cancel();
         }
+
+        client->ActiveClientCalls.clear();
```
shivaprasad-basavaraj, line 239: Is `clear()` atomic? Should we get a lock here as well before we clear the list?

shivaprasad-basavaraj (approved):

src/grpc_client.cc
ni-sujain: yes, this is a scoped lock

ni-sujain:
Note: This PR still need some work. There are test failures for streaming cases

ni-sujain:
See PR https://github.com/ni/grpc-labview/pull/251 for this closing this

PR #229, opened 2023-02-02
bump vipb version to 1.0.0.1

ni-sujain:
Bumping versions for .vipb files to 1.0.0.1(format major.minor.patch.build). For .vipb files build number cannot be zero


ni-sujain:
will skip the CI as workflow is broken

PR #234, opened 2023-03-03
Set connector pane terminals for Main.vi

pratheekshasn:
The connector pane for the Main.vi that generates the code for a given proto file.

image

This change is necessary for adding tests.


ni-sujain (commented):
You can mention in description that for calling into tests we are creating connector pane

ni-sujain (approved):

ni-sujain:
Since this does not have any code changes that require creating binaries and passing tests, willl do force merge

Issue #228, opened 2023-02-02
Run CI tests action times out


ni-sujain:
when runtest.py script kicks in it executes the first test and after that goes into an indefinite wait, finally it times out. 

AB#2288154

ni-sujain:
Debugging note: Seems like either a problem with python or LabVIEWCLI. The subprocess run is stuck into wait. If I run run_tests.py locally it passes but times out when invoked from workflow

PR #239, opened 2023-03-30
fix hang problem with python subprocess run while running tests

ni-sujain:
fix for issue https://github.com/ni/grpc-labview/issues/228
While running tests for grpc-labview, we use subprocess.run() python method in run_tests.py. It used to hang randomly after running one or two tests. I found that there are open issues([75628](https://github.com/python/cpython/issues/75628), [76976](https://github.com/python/cpython/issues/76976)) wrt. python subprocess.run() method involving pipes and timeouts. For now a neat way to do this is to use Popen.communicate() as suggested in documentation. For now we are keeping the timeout for a test run as 60 secs if the process doesnot terminate in that time it gets handled via TimeoutExpiredException and gets killed so that we do not go into a hang state


ni-sujain (commented):
need to re-visit 60 sec timeout, it might not be enough

PR #240, opened 2023-04-03
Create enum mappers between LV enums and proto values

pratheekshasn:
We don't have a one-to-one mapping between the enums created on the LabVIEW side and the enum defined in the proto file, considering that enums in a proto file can have non-sequential, negative and even non-unique values.

To allow the users to access the underlying numeric value in an enum (as defined either in the proto file or on the LabVIEW side), mapper VIs have been created.

### Template for mapper VIs
Both of the VIs below are added to Client Template.lvlib and Server Template.lvlib
**Get package_template_enum for Value.vi**
image

**Get Value for package_template_enum.vi**
image

### Scripting code for creating mapper VIs

**Create Enum Data Clusters (simple).vi**
image

**Create Mapper VIs for Enum.vi**
image

**Create Enum Mapper.vi**
image

**Update Enum Mapper.vi**
image

### Delete scripting template files

**Main.vi**
image

**Delete Message Template Files.vi**
image

### Refactor changes

**Create Message Getter (from Template).vi** has been changed so it can be reused to script out enum mappers.
image

This required a change in the caller of the subVI, as shown below:
**Create Message Getter.vi**
image


src/grpc_interop.cc
```diff
@@ -285,6 +285,36 @@ LIBRARY_EXPORT int32_t RegisterEnumMetadata2(grpc_labview::gRPCid** id, grpc_lab
...
+{
+    auto server = (*id)->CastTo<grpc_labview::MessageElementMetadataOwner>();
+    if (server == nullptr)
+    {
+        return -1;
+    }
+    auto metadata = (server.get())->FindEnumMetadata(std::string(enumName));
+    return metadata.get()->GetLVEnumValueFromProtoValue(protoValue);
+    
+    //return 0;
```
ni-sujain, line 300: remove comment, as discussed use the error code as return type. for proto value pass int*, on the caller side user error handler subvi.

src/grpc_interop.cc
```diff
@@ -285,6 +285,36 @@ LIBRARY_EXPORT int32_t RegisterEnumMetadata2(grpc_labview::gRPCid** id, grpc_lab
...
+{
+    auto server = (*id)->CastTo<grpc_labview::MessageElementMetadataOwner>();
+    if (server == nullptr)
+    {
+        return -1;
+    }
+    auto metadata = (server.get())->FindEnumMetadata(std::string(enumName));
+    return metadata.get()->GetProtoValueFromLVEnumValue(lvEnumValue);
+
+    //return 0;
```
ni-sujain, line 315: see above comment

ni-sujain (changes requested):
needs changes in the new api signature

src/grpc_interop.cc
pratheekshasn: Fixed.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Enum Mapper.vi
ni-sujain, line 1: whats the nre change here

src/grpc_interop.cc
```diff
@@ -285,6 +285,38 @@ LIBRARY_EXPORT int32_t RegisterEnumMetadata2(grpc_labview::gRPCid** id, grpc_lab
     return 0;
 }
 
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+LIBRARY_EXPORT int32_t GetLVEnumValueFromProtoValue(grpc_labview::gRPCid** id, const char* enumName, int protoValue, int* lvEnumValue)
```
ni-sujain, line 290: should this be uint* instead of int*?

ni-sujain (approved):
consider changing the api signature to return uint* which gives us lv enum index

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Enum Mapper.vi
pratheekshasn: I had moved a wire so that its label would be visible.

src/grpc_interop.cc
pratheekshasn: Fixed.

Issue #243, opened 2023-04-09
Errors while trying to regenerate code after making some modifications in the messages in the same protofile.


yash3001:
Labview's gRPC template creation utility throws the following error if I generate the code once and then try to add a new message in the protofile and regenerate the code.

**- Error:**

   ![230800516-235fa785-f2ad-4cf7-aee7-5d4c63850ea5 (1)](https://user-images.githubusercontent.com/43678393/230800570-d96f294c-5736-4cf7-82d7-ecbfc9c5df18.png)

**-  Software Environment**

1. Windows 10
2. Labview 2022Q3 (22.3.1f8)

**- You can reproduce the error by following the below steps:**

1. Create a protofile "greet.proto" with the following code:
    ```protobuf
    syntax = "proto3";
    
    package newgreeter;
    
    service Greeter {
        rpc SayHello (HelloRequest) returns (HelloReply);
        rpc ParrotSayHello (HelloRequest) returns (stream HelloReply);
        rpc ChattyClientSayHello (stream HelloRequest) returns (DelayedReply);
        rpc InteractingHello (stream HelloRequest) returns (stream HelloReply);
    }
    
    message HelloRequest{
        string message = 1;
    }
    
    message DelayedReply{
        string message = 1;
        repeated HelloRequest request = 2;
    }
    
    message HelloReply{
        string message = 1;
    }
    ```

2. Create a new blank labview project "project.lvproj" and configure the code generator to use the newly created labview project and the above protofile. Run the code generator.

    ![230800095-bed3e5ce-88d6-40d2-8909-d17ef3e6f263](https://user-images.githubusercontent.com/43678393/230800374-61cfd26d-f081-46d7-82b8-17083b4b0c1b.png)

3. Now modify the protofile by adding a new message in the end as follows:
    ```protobuf
    syntax = "proto3";
    
    package newgreeter;
    
    service Greeter {
        rpc SayHello (HelloRequest) returns (HelloReply);
        rpc ParrotSayHello (HelloRequest) returns (stream HelloReply);
        rpc ChattyClientSayHello (stream HelloRequest) returns (DelayedReply);
        rpc InteractingHello (stream HelloRequest) returns (stream HelloReply);
    }
    
    message HelloRequest{
        string message = 1;
    }
    
    message DelayedReply{
        string message = 1;
        repeated HelloRequest request = 2;
    }
    
    message HelloReply{
        string message = 1;
    }
    
    message NewMessage{
        string message = 1;
    }
    ```

[...]

PR #245, opened 2023-04-25
Fixed an error that was thrown upon regeneration of server code after adding a new message in the protofile. 

yash3001:
Fix for the issue #243.

#### This PR contains the following change to "Create Message Getter (from accessor).vi" :

* #### "Create Message Getter (from accessor).vi" before change
  
     pre

* #### "Create Message Getter (from accessor).vi" after change
    
    after

#### NOTE :
1) After this change, the error is no longer thrown and all the vi's are generated for the newly added message.
2) However this may not be the ideal behavior since there is no way to differentiate between a newly added message and a message that was already present but whose name is now changed.
3) Also, after regeneration, the message specific vi's (getters, setters, controls) are not deleted for a message that is removed from the protofile before running the code generator. They should be removed in the ideal case.


ni-sujain:
see PR https://github.com/ni/grpc-labview/pull/250

PR #250, opened 2023-05-02
Fixed an error that was thrown upon regeneration of server code after adding a new message in the protofile.

yash-ni:
Fix for the issue #243.

#### This PR contains the following change to "Create Message Getter (from accessor).vi" :

* #### "Create Message Getter (from accessor).vi" before change
  
     pre

* #### "Create Message Getter (from accessor).vi" after change
    
    after

#### NOTE :
1) After this change, the error is no longer thrown and all the vi's are generated for the newly added message.
2) However this may not be the ideal behavior since there is no way to differentiate between a newly added message and a message that was already present but whose name is now changed.
3) Also, after regeneration, the message specific vi's (getters, setters, controls) are not deleted for the message that was removed from the protofile before running the code generator. They should be removed in the ideal case.


labview source/Client Server Support New/build spec/gRPC Server and Client Template [2].vipb
Comment (ni-sujain) [inline labview source/Client Server Support New/build spec/gRPC Server and Client Template [2].vipb:1]: revert this file

ni-sujain (approved):
just resolve conflicts

PR #262, opened 2023-05-17
[Crash Fix] Shutdown and drain completion queue when stopping server

gaoruhao:
# Justification
A crash was found when ADAS and VeriStand team were trying to debug a bug in EVO engine. The crash occured at `GPR_ASSERT(queue.num_items() == 0);` as shown below:

**completion_queue.cc**
```c++
struct cq_next_data {
  ~cq_next_data() {
    GPR_ASSERT(queue.num_items() == 0);
#ifndef NDEBUG
    if (pending_events.load(std::memory_order_acquire) != 0) {
      gpr_log(GPR_ERROR, "Destroying CQ without draining it fully.");
    }
#endif
  }
```

According to this stackoverflow [thread](https://stackoverflow.com/questions/35708348/grpc-what-is-the-recommended-way-to-shut-down-an-asynchronous-server-in-c), we should also shutdown the completion queue during server shutdown. It is very likely to be the root cause of the crash.

See more details here:
https://ni.visualstudio.com/DevCentral/_workitems/edit/2326901/

# Implementation
My fix is simple, which is to shutdown the completion queue during server shutdown. In the meanwhile, we also need to drain the completion queue for the existing items. In our current implementation, the server listening thread will be stopped immediately so it does not have chances to drain the completion queue in the listening thread.

# Test
Tested locally on PXI RT by using the test project attached in the Azdo bug.
https://ni.visualstudio.com/DevCentral/_workitems/edit/2326901/


src/grpc_server.cc
```diff
@@ -139,15 +142,16 @@ namespace grpc_labview
         bool ok;
         while (true)
         {
+            if (_shutdown)
```
gaoruhao, line 145: Move the flag check before checking the completion because it is possible that users tries to close the server very quickly so when we enter this while loop, the server and completion queue are already shutdown.

src/grpc_server.cc
```diff
@@ -237,9 +241,28 @@ namespace grpc_labview
         if (_server != nullptr)
         {
             // We need shutdown passing a deadline so that any RPC calls in progress are terminated as well.
-            _server->Shutdown(std::chrono::system_clock::now());
-            _server->Wait();
-            _runThread->join();
+            _server->Shutdown();
```
ni-sujain, line 244: Some questions:
a) why did we remove the deadline for server->shutdown() call here?
b) aren't we suppose to wait for server->shutdown() to complete

ni-sujain (changes requested):
@gaoruhao I have some questions.

src/grpc_server.cc
gaoruhao: You are right. I should not modify this.

gaoruhao: The code is updated.

ni-sujain (commented):
I agree that we should do this part of draining the completion queue but should we do it before or after thread join I am not sure? what are your thoughts?

src/grpc_server.cc
```diff
@@ -239,7 +243,27 @@ namespace grpc_labview
...
+            {
+                _cq->Shutdown();
+            }
+
+            if (_runThread->joinable())
+            {
+                _runThread->join();
+            }
+
+            // Drain the complete queue before deleting the server.
```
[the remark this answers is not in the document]
gaoruhao: Based on my test, the drain code needs to be here. I didn't dig into more details.

ni-sujain (approved):

PR #263, opened 2023-05-19
[US2366055] Update "LabVIEW DataType to Message Element DataType.vi" to use gRPC "Enum Value"

WesleyTangNationalInstruments:
Updated `LabVIEW DataType to Message Element DataType.vi` to convert `Eum U8`, `Enum U16`, `Enum U32`, and `Enum U64` to use gRPC `EnumValue` Message Element DataType instead. Previously, the were converted to their corresponding `UInt` type.

image

[AB#2366055](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2366055)


JasonC512:
Where and how is this VI used? Are all of the callers in the other PR you created? ie: [https://github.com/ni/measurementlink-labview/pull/213](https://github.com/ni/measurementlink-labview/pull/213)

The diff looks reasonable, but it is difficult to know if this would break any assumptions made by callers. Will caller VIs break with a compiler error because of this change? Or would it just be potential logic errors?

WesleyTangNationalInstruments:
@JasonC512 

> Where and how is this VI used? 

This VI is called by `MeasurementLink Measurement Server.lvlib:Get MetaData from Configuration.vi` and `MeasurementLink Measurement Server.lvlib:Get MetaData from Result.vi` to populate the configuration and output parameters. 

See usage:
image

> Are all of the callers in the other PR you created? ie: https://github.com/ni/measurementlink-labview/pull/213 

The callers of this VI already exist in the MeasurementLink repo. My other PR adds no new callers.

> The diff looks reasonable, but it is difficult to know if this would break any assumptions made by callers. Will caller VIs break with a compiler error because of this change? Or would it just be potential logic errors?

Based on previous discussions with Everett, we think that this should be pretty safe. The existing behavior is that LabVIEW enums get coerced to gRPC UInts and there is no LabVIEW types that converts to gRPC Enums. This change results in LabVIEW enums now being converting to gRPC Enums instead of gRPC UInts. The main issue I could see is that if an existing caller relies on LabVIEW enums being coerced to gRPC UInts they would need to update their handling, but I don't think there are any callers trying to use other than MeasurementLink. This would be more of a change in behavior rather and I'd lean towards this unlikely to cause compile errors.

ni-sujain (approved):

Issue #274, opened 2023-07-17
LabVIEW DataType to Message Element DataType.vi is broken


pbirkhol-ni:
![image](https://github.com/ni/grpc-labview/assets/57813166/98b91696-16c8-448c-9d2a-af4f10dd4047)
It doesn't wire the `Supported?` output of the case structure for the `Enum U8..Enum U64` case.

PR #283, opened 2023-08-07
Fixed broken DataType to Message Element DataType.vi

yash-ni:
Fix for issue #274 

Fixed the vi by wiring `True` to the `Supported?` output of the case structure for the `Enum U8..Enum U64` case.

image


ni-sujain:
@yash-ni  Just make sure that VI is saved in LV 2019 version

PR #294, opened 2023-08-17
Implement ListServices in reflection

pratheekshasn:
ListServices() had not been implemented for gRPC reflection. This PR adds support for the same.


CPattar-NI (approved):

src/lv_proto_server_reflection_service.cc
```diff
@@ -80,16 +80,21 @@ namespace grpc_labview
 
 
     void LVProtoServerReflectionService::AddFileDescriptorProto(const std::string& serializedProtoStr) {
-        // reflection_service_.get()->AddFileDescriptorProto(serializedProto); 
         FileDescriptorProto proto;
         proto.ParseFromString(serializedProtoStr);
-        other_pool.BuildFile(proto);
+        other_pool_file_descriptor = other_pool.BuildFile(proto);
```
ni-sujain, line 85: As an alternative instead of storing a `grpc::protobuf::FileDescriptor* other_pool_file_descriptor` in the `ReflectionService`, we can return a  `grpc::protobuf::FileDescriptor*`  from this method `AddFileDescriptorProto`.

In the caller of this method, we can move the logic (from line 92-97) to  iterate over the Descriptors and call `AddService `method. That should make it clean and we don't have to store the `other_pool_file_descriptor`

src/lv_proto_server_reflection_service.cc
```diff
@@ -80,16 +80,21 @@ namespace grpc_labview
...
     }
 
 
     Status LVProtoServerReflectionService::ListService(ServerContext* context,
         grpc::reflection::v1alpha::ListServiceResponse* response) {
 
+        int numServices = other_pool_file_descriptor->service_count();
+        for (int i = 0; i < numServices; ++i) {
+            const google::protobuf::ServiceDescriptor* serviceDescriptor = other_pool_file_descriptor->service(i);
```
ni-sujain, line 94: see comment above

ni-sujain (approved):

src/lv_proto_server_reflection_service.cc
pratheekshasn: Fixed.

src/lv_proto_server_reflection_service.cc
pratheekshasn: Fixed.

src/lv_proto_server_reflection_service.cc
chethanravindranath: @pratheekshasn @ni-sujain, calling AddService is not a good idea as the AddService modifies the plugin service too. We need to keep the FileDescriptor and DescriptorProto separate from the plugin object itself. Modifying the state of the reflection service itself might have unintended and unknown consequences. For e.g., if I run example-one server followed by example-two, we end up listing example-one's services when listing the example-two's services. This is incorrect. I believe that adding the FileDescriptor* is the right way to go about the fix. Can you please revert this fix?

chethanravindranath (approved):

Issue #308, opened 2023-09-11
Support getting enum metadata from a LabVIEW cluster


pbirkhol-ni:
When using an enum message, grpc-labview requires you to register enum metadata. This works fine when generating a client/server using grpc-labview's code generator, but MeasurementLink doesn't do this. Instead, MeasurementLink uses `Cluster to gRPC LabVIEW Message.vi` to dynamically generate message metadata at runtime. This has always worked fine, but with grpc-labview 1.0.0.7, it no longer works with enums. The reason is that we now have to register enum metadata with grpc-labview.

One possible fix is to add a VI to generate the enum metadata given a cluster, similar to how `Cluster to gRPC LabVIEW Message.vi` works. Then MeasurmentLink would call that new VI and register the resulting metadata.

AB#2518465

pbirkhol-ni:
I have verified in the latest release that this is fixed. The overall issue that measurementlink-labview is trying to solve is not yet fixed, but this issue as written has been verified.

Issue #295, opened 2023-08-17
Multiple Enums Copying Incorrectly to Cluster


WesleyTangNationalInstruments:
This bug was found in InstrumentStudio using MeasurementLink LabVIEW where multiple consecutive enums were not being sent from the client (InstrumentStudio) to the server (MeasurementLink LabVIEW measurement) correctly. Debugging into this, it seems that the issue lies somewhere grpc-labview where the only the first enum's values get transcribed into the cluster structure. Sumit Jain (@ni-sujain) asked me to recreate this bug item directly in the grpc-labview repo.

Further details and an example measurement that demonstrates this issue can be found in this [AzDO bug item](https://ni.visualstudio.com/DevCentral/_workitems/edit/2478685). 

For completeness, I've also detailed the issue and debugging steps taken below.

------------------------------------------------------

## Bug Description

This behavior was reported by Andy Tan Wei Keat (andy.wei.keat.tan@ni.com) using MeasurementLink LabVIEW v1.2.0.2 which is built against grpc-labview v1.0.0.1. An example measurement was provided to reproduce this issue and attached. 

[Sample Plugin 1001.zip](https://github.com/ni/grpc-labview/files/12372331/Sample.Plugin.1001.zip)

Note that MeasurementLink LabVIEW is currently incompatible with the latest version of grpc-labview and cannot be upgraded from v1.0.0.1.

This example sets up two enums where the configuration pipes directly into the result.

![673cbc01-fdd5-4277-ae8b-8826d8a47dfb](https://github.com/ni/grpc-labview/assets/103216564/e5819053-43ba-4b0c-b35c-2c68685235f1)

When we have two continuous enums in a Measurement Configuration, the second enum will not receive the data in the Measurement Result.

![3d6a27ee-d394-4647-b6d1-a0d8915b42fc](https://github.com/ni/grpc-labview/assets/103216564/9ca576e1-506e-47f5-a97b-f28b8bd8da16)

Note that if a non-enum control is inserted in between the two enum controls, the data is passed as expected.

![2f23b91d-e424-439f-bfdb-7c05a4bb4a00](https://github.com/ni/grpc-labview/assets/103216564/3010826b-e48a-4f62-9ac0-507063718917)

![92b83e06-8d82-438f-a535-e5dff736a715](https://github.com/ni/grpc-labview/assets/103216564/542b9578-413d-40a0-a3d7-c1d4aa214583)

------------------------------------------------------

## Debugging Steps Taken

[...]

PR #309, opened 2023-09-11
Add support for getting enum metadata from cluster

pbirkhol-ni:
This is a possible fix for #308.

This change adds a new output to `Cluster to gRPC LabVIEW Message.vi`. The new output is an array of `Enum Metadata.ctl` that can be registered to `grpc-labview`.

Why add it to `Cluster to gRPC LabVIEW Message.vi` instead of creating a new VI to get the enum metadata? Two reasons:
- `Cluster to gRPC LabVIEW Message.vi` already recursively iterates over the cluster, so it is more performant to only do it once.
- Any client who is calling `Cluster to gRPC LabVIEW Message.vi` will almost certainly want the enum metadata anyway.
 
A few questions:
- Should we check for duplicate enum definitions? What would happen if the same enum got registered multiple times?
- How do I get the linking correct between `grpc-lvsupport-release.lvlib` and `grpc-lvsupport.lvlib`? While testing this, I copied the new VIs to vi.lib and relinked to the `grpc-lvsupport-release.lvlib` library.


pratheekshasn:
> Should we check for duplicate enum definitions? What would happen if the same enum got registered multiple times?

How would you duplicate enums in the first place? This is what I understand happens with the MeasurementLink code, and I base my answer off of that, please correct me where I'm wrong. (I inferred this from the project attached in #295.)

1. The project has ctls, which I'm assuming corresponds to the messages.
2. The Run Service VI registers these messages by:

- Getting reference of each of these CTLs.
- Getting the control's default value out as a variant.
- Passing the variant to Cluster to Element Details.vi which creates the message metadata and the enum metadata out of this variant.

Then this newly created metadata is sent to RegisterMessageMetadata2.

Since Run Service iterates through all the CTLs, why would there be duplicate enums?

PS: Even if yo do somehow end up registering enums twice, since the metadata is stored in a map, any existing key will not be updated the second time onwards.


> How do I get the linking correct between grpc-lvsupport-release.lvlib and grpc-lvsupport.lvlib? While testing this, I copied the new VIs to vi.lib and relinked to the grpc-lvsupport-release.lvlib library.

That should be fine as long as no VI in grpc-lvsupport.lvlib links to another VI in grpc-lvsupport.lvlib, and still maintains links to grpc-lvsupport-release.lvlib.

pratheekshasn (approved):

pbirkhol-ni:
> Since Run Service iterates through all the CTLs, why would there be duplicate enums?

The measurement data could have the same enum two times. That enum could even have the same label, which LabVIEW allows. However, MeasurementLink doesn't allow two controls with the same label and will throw a helpful error if you try to do this. I wanted to make sure that grpc-labview doesn't override our error and my testing indicates that it doesn't. grpc-labview is happy to register the same enum metadata mulitple times.

Another reason why I wanted to check is because `Cluster to gRPC LabVIEW Message.vi` is a generic VI. It should work for all use cases, not just MeasurementLink's use case. 

> PS: Even if yo do somehow end up registering enums twice, since the metadata is stored in a map, any existing key will not be updated the second time onwards.

This eases my concerns. It would still lead to inconsistent behavior if you tried to register two different enums with the same label. This is not a concern for MeasurementLink due to how we use enums. It might be a concern for other clients of `Cluster to gRPC LabVIEW Message.vi`.

Issue #265, opened 2023-05-30
Event structure in 'Start Sync.vi' is not updating when we try to regenerate after adding a new rpc method in protofile


yash-ni:
When I try to add a new rpc method under a service in the protofile and then regenerate the server code, then the event structure (in Start Sync.vi) is not updating to show the newly added rpc method's event in it.

### Software Environment:
1. Windows 10
2. Labview 2023 Q1 (64 bit)

### Original Protofile "greet.proto" (before modifications) that I used to generate the code for the first time:

```proto
syntax = "proto3";

package newgreeter;

service Greeter {
    rpc SayHello (HelloRequest) returns (HelloReply);
    rpc ParrotSayHello (HelloRequest) returns (stream HelloReply);
    rpc ChattyClientSayHello (stream HelloRequest) returns (DelayedReply);
    rpc InteractingHello (stream HelloRequest) returns (stream HelloReply);
}

message HelloRequest{
    string message = 1;
}

message DelayedReply{
    string message = 1;
    repeated HelloRequest request = 2;
}

message HelloReply{
    string message = 1;
}
```

### Code Generator configurations (will remain same):

![generator](https://user-images.githubusercontent.com/43678393/231116557-402d8adc-96e3-4ffd-8387-cc32540d6107.png)

### Steps to reproduce the error:

1. Generate the code for the first time using the above protofile and generator configuration.

2. You will notice the following events in the event structure in 'Start Sync.vi' 
   
    ![event structure](https://github.com/ni/grpc-labview/assets/131851960/cc1707c3-cb51-477e-84c7-51ebdbf20888)

3. Now update the protofile by adding a new rpc method as given below:
    ```proto
    syntax = "proto3";
    
    package newgreeter;
    
    service Greeter {
        rpc SayHello (HelloRequest) returns (HelloReply);
        rpc ParrotSayHello (HelloRequest) returns (stream HelloReply);
        rpc ChattyClientSayHello (stream HelloRequest) returns (DelayedReply);
        rpc InteractingHello (stream HelloRequest) returns (stream HelloReply);
        rpc NewMethod (HelloRequest) returns (HelloReply);
    }
    
    message HelloRequest{
        string message = 1;
    }
    
    message DelayedReply{
        string message = 1;
        repeated HelloRequest request = 2;
    }
    
    message HelloReply{
        string message = 1;
    }
    ```

4. Use the generator to generate server side code.

5. You can see that the event structure is not updating and not adding the new method's event.

[...]

PR #314, opened 2023-09-21
Fixed issue with event structure not upgrading in 'start sync'.vi

yash-ni:
Fix for bug #265


labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Generate RPC Methods.vi
ni-sujain, line 1: change the type to int64 for newly added method count

ni-sujain (approved):
try some tests locally against the proto file changes

Issue #313, opened 2023-09-21
Scripting multiple protos that call the same sub-proto: The sub-proto will be created multiple times.


SallyTYT:
I have multiple protos, all of which have called [`/google/protobuf/descriptor.proto`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto). When I script these protos separately, it works. When I script more than one proto together, it will report an error as shown below.

![image](https://github.com/ni/grpc-labview/assets/64485819/a62f22f3-99ab-428d-8421-0cf239e07263)

 I checked the uncompleted codes, and found it seems that the `/google/protobuf/descriptor.proto` was created multiple times, causing the enum ctrl to be broken as below,

![image](https://github.com/ni/grpc-labview/assets/64485819/052b7925-4a81-4957-a7bd-aa7d81c277dc)

Workaround: 
Scripting these protos separately, although the sub-proto will have multiple copies.

SallyTYT:
It seems that inserting "Remove Duplicates From 1D Array.vim" here could solve this issue.

![image](https://github.com/ni/grpc-labview/assets/64485819/6f8d1621-3063-4cf5-b689-6c5b7405f163)

jasonmreding:
Appears to be duplicate of #340

Issue #186, opened 2022-11-08
Code generator Import Paths are asking for .proto files instead of folders


dixonjoel:
We have a root Protos directory and all imports in .proto files are relative to that, so we'd like to add folders that search for .proto files recursively to the Import paths. I think this is the intention of the Import Paths in the code generator, but their browse options are set for *.proto files instead of folders. If I change these browse options and pick a path, it works correctly.
![image](https://user-images.githubusercontent.com/38357562/200689990-648ca246-8845-41ef-ac04-63feb5c7ca0a.png)

Issue #38, opened 2021-11-18
Separate Compiled Code missing for some VIs


chrisb2244:
Some VIs do not have "Separate compiled code from source file" checked, which leads to the files being modified when opened on a different platform (I guess using 32-bit LabVIEW is not the norm here?).

I could easily change these and mark them all, but that will probably cause merge conflicts since the files might be in various unrelated places and then annoy everyone else trying to work in this repository :/

I'll add the checkbox to any files that I actually modify, but I'm opening the issue in case this repository has a period of time (e.g. releases?) when everyone is likely to not have open branches, in which case perhaps the files could be mass-ticked?

Alternatively, is there some preferred strategy for handling this? (Or are the files intentionally not separated from compiled code?)

navinsubramani:
@chrisb2244  I guess we all wanted to have the check box enabled for all VIs. I like your point about making the change when we want to change that VI for other purposes, and before the next release maybe @ccifra does the check for all. 

Yes, Merging is a pain!

jasonmreding:
Aside from testing related VIs, this should be fixed. The project used for editing the source is configured so that new project items will default to this setting and all existing items in the project have also been configured so that "Separate compiled code from source file" is checked.

PR #325, opened 2023-12-01
Fix browse for proto imports and other minor cleanup

jasonmreding:
This addresses #186. It also includes changes related to #313 and #38.


labview source/Client Server Support New/gRPC Scripting Tools/Main.vi
jasonmreding, line 1: The only change here is hidden in right click configuration options of the path control(s) used to specify directories from which to find imported proto files. The path control is now configured to browse for directories instead of files.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Rich to Flat Conversion.vi
jasonmreding, line 1: The only change here is to separate compiled code from source.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Wire Between Unbundle And Bundle For Message Converter With Oneof.vi
jasonmreding, line 1: The only change here is to separate compiled code from source.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Create Oneof Message Converters (Rich to Flat).vi
jasonmreding, line 1: I started making changes to this and `Create Oneof Message Converters.vi` as I was debugging and trying to improve code generation behavior when regenerating code in project in which previously generated code already exists. However, I soon realized making this developer flow work well probably requires an extensive overhaul of the existing code, which is more than I have time for at the moment. However, I decided to submit these VIs anyway since they have been simplified quite a bit, and are much easier to read.

labview source/Client Server Support New/gRPC Scripting Tools/Project API/Delete In Folder.vi
jasonmreding, line 1: This VI didn't think it belonged to the library for some reason. Now it does.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Enum Data Clusters (simple).vi
jasonmreding, line 1: I also modified this while trying to improve behavior of regenerating a proto file without first deleting everything that was previously generated. This change prevents multiple copies of the top level enum from getting added to the typedef, which breaks the typedef. Having a broken typedef in turn generates lots of downstream errors with VI scripting when adding things by path and what not.

jasonmreding:
@pratheekshasn Can you review this PR and/or assign to appropriate reviewers. It appears I don't have write access so I can't add reviewers myself. Would it also be possible for me to get write access to the repo?

pratheekshasn (approved):

PR #338, opened 2024-01-11
Feature Toggles and Debug Symbols for Release Build

CPattar-NI:
Adding infrastructure to facilitate feature toggles. This will be a singleton class that reads from a config INI file, enabling and disabling a given feature.

Enabling the creation and storage of symbol files for release builds.


pratheekshasn (approved):

src/feature_toggles.cc
```diff
@@ -0,0 +1,49 @@
+#include "feature_toggles.h"
+#include <iostream>
+#include <fstream>
+#include <sstream>
+#include <map>
+
+// Function to read feature configurations from an INI file
+void FeatureConfig::readConfigFromFile(const std::string& filePath) {
```
ni-sujain, line 8: Consider adding some documentation for the format of the ini file:

[section1]
key1=value1
key2=value2
[section2]
...

src/grpc_client.cc
```diff
@@ -276,6 +277,9 @@ LIBRARY_EXPORT int32_t ClientUnaryCall(
     int32_t timeoutMs,
     grpc_labview::gRPCid* contextId)
 {
+    // Instantiating the feature toggles singleton that will read the feature configuration file
+    FeatureConfig::getInstance().readConfigFromFile("feature_config.ini");
```
ni-sujain, line 281: Since we have already called this in InitCallbacks() do we need this here.

src/grpc_client.cc
```diff
@@ -401,6 +405,9 @@ LIBRARY_EXPORT int32_t ClientBeginClientStreamingCall(
...
```
ni-sujain, line 409: same comment as above

src/lv_interop.cc
```diff
@@ -51,6 +52,9 @@ namespace grpc_labview
     //---------------------------------------------------------------------
     void InitCallbacks()
     {
+        // Instantiating the feature toggles singleton that will read the feature configuration file
+        FeatureConfig::getInstance().readConfigFromFile("feature_config.ini");
```
ni-sujain, line 56: This seems to be a correct place to call readConfig . InitCallbacks() is the first method to get called while we do CreateClient, CreateServer, CreateSerializationSression etc.

src/grpc_server.cc
```diff
@@ -178,6 +179,9 @@ namespace grpc_labview
         std::string serverKeyPath,
         ServerStartEventData *serverStarted)
     {
+        // Instantiating the feature toggles singleton that will read the feature configuration file
+        FeatureConfig::getInstance().readConfigFromFile("feature_config.ini");
```
ni-sujain, line 183: same comment as above

ni-sujain (approved):

src/feature_toggles.cc
CPattar-NI: Isn't this the generic INI format?

ni-sujain: yes

src/lv_interop.cc
CPattar-NI: Removed it in all other locations.

Issue #345, opened 2024-02-05
VI Hangs when multiple client reading stream VIs are waiting and multiple Steam Ends occur. This used to work in 1.0.0.1.


NIbbuchanan:
Our application used to work with grpc 1.0.0.1, but I discovered there was a change between 1.0.0.3 and 1.0.0.4 that broke our application and caused it to hang. We are now using 1.2.0.1 (latest version available), which also results in a hang as well. When there are multiple client stream reads waiting for a message and the client streams all end in parallel, some of the reads never return. They used to all return when the streams were ended with either an error or the valid flag being false, but now a simple test VI that reproduces this issue just hangs indefinitely. A work around is to avoid reading in parallel and let each one fail/timeout sequentially when the streams are closed, but in some cases, the reads are spread across various locations so it's not trivial to coordinate when they run. I'd like to get a fix so the previous behavior works since a hang seems bad. Let me know if you need any additional information. Thanks!

[GrpcHangBug.zip](https://github.com/ni/grpc-labview/files/14171809/GrpcHangBug.zip)

AB#2645647

nischalks:
@NIbbuchanan I would suggest you to move to the latest release https://github.com/ni/grpc-labview/releases/tag/v1.2.0.1 and verify if it solves your problem. Thanks!

NIbbuchanan:
I updated the original message because I forgot to mention we are using the latest 1.2.0.1, which is when I discovered this issue after we upgraded from 1.0.0.1. I installed older versions to determine when it broke to hopefully make it easier to track down the change that broke this behavior.

nischalks:
@NIbbuchanan This is now fixed and released as part of https://github.com/ni/grpc-labview/releases/tag/v1.2.1.1

NIbbuchanan:
I agree the latest export fixes this issue....Thanks! I did find a new issue with the vi.lib\gRPC\LabVIEW gRPC Library\Shared\Wait On Occurrence Wrapper.vi on the latest export:
- It is not marked as re-entrant/in place so if a grpc call takes awhile, it blocks other calls from going through.
- The separate compiled code flag is not checked, which is annoying when going between different contexts (ie. windows an RT) and this VI has to be saved each time. I'll follow a separate issue for this...it was causing our app to hang until I found this problem.

NIbbuchanan:
[...]

Issue #330, opened 2023-12-07
Calling gRPC Client VIs from TestStand hang at "Wait for Occurrence" while using LabVIEW RTE


kt-jplotzke:
I've found a condition where calling gRPC client VIs from TestStand, while using the LabVIEW Run Time Engine adapter, causes the "Client Unary Call" VI to hang at "Wait on Occurrence".  I can confirm that the gRPC request _does_ get sent from the client prior to hanging and that the server _does_ respond to the request.  This issue does not occur when running using the LabVIEW Development Environment adapter.

Steps that I've used to create a reproducible example are below.  I've also uploaded all of this code to [https://github.com/kt-jplotzke/grpc-labview-teststand-hang](https://github.com/kt-jplotzke/grpc-labview-teststand-hang).

I've created a small gRPC service with a single function, "Wait".  This function simply wraps the LabVIEW Wait (ms) function.  I then used grpc-labview 1.0.1.1 to generate the server and client code for this service.

```
service TestService {
	rpc WaitMs(WaitRequest) returns (WaitResponse);
}

message WaitRequest {
	uint64 msec_to_wait = 1;
}

message WaitResponse {
	uint64 timestamp = 1;
}
```

For the server, I simply wrapped the Wait (ms) function in the "Start Sync" function.  This is the only gRPC method.  I then built `Run Service.vi`  as an EXE to allow this to run in the background.
![image](https://github.com/ni/grpc-labview/assets/88543727/d818f840-1cdc-46b8-92a5-f0f368175a9d)

For the client, I simply wrapped my gRPC client function into a VI which calls `Client Create`, the gRPC client unary function, and then `Client Destroy`.  I build this function into a Packed Project Library.
![image](https://github.com/ni/grpc-labview/assets/88543727/9cb47b73-ced7-44fe-8a49-ff7dce1f0111)

I can confirm that both the native wrapper VI and the build PPL wrapper VI function as expected when executed from the LabVIEW Dev Environment.  Both send a request and receive a response from the built gRPC server.

Then, I created a TestStand sequence with a single step -- to call this gRPC client wrapper VI.
![image](https://github.com/ni/grpc-labview/assets/88543727/14ff4335-962d-4458-a316-c4877b59e097)

[...]

Issue #324, opened 2023-11-09
grpc-labview client call hangs even though the server responded.


rdecarreau:
We have an application in which we can replicate a client-side hang of the grpc-labview binary. We are almost positive (still gathering data) that the server replied, and we have configured the grpc-labview timeout on the client. What can we do to gather information while the hang is happening? What can I do to help someone take a look?

AB#2576440

rdecarreau:
LabVIEW 2023 Q3
cRIO RT Target on 23Q3
grpc-labview 1.0.0.5 and 1.0.0.8

rdecarreau:
I have a VI which can replicate but since we have to run it in our lab, I will need to work with someone who can help.

AndrewHeim:
This sounds similar to my problem I had today - #193 was behind it.  Try ruling that out and see if it helps?

I don't have a LV 2019 environment to do it with right now, or I'd already have submitted the pull request.

rdecarreau:
Hahahah. That was our team. Wow. Yes very much related and is what I did as a workaround.

nischalks:
This is now fixed in this release of https://github.com/ni/grpc-labview/releases/tag/v1.2.1.1, please confirm if the hang goes away with this release.

Issue #320, opened 2023-10-09
Calling gRPC client API in VMWare machine will sometimes hang


wendisong:
Hi,

Currently I am developing my code in VMWare machine, and I also use grpc as dependency. I found that if I try to call generated grpc client API for overnight test, then it will hang.
I attached my test code here.
[gRPC_Hang_Issue.zip](https://github.com/ni/grpc-labview/files/12844113/gRPC_Hang_Issue.zip)

### VMWare Machine
CPU: i7-6700 2C4T
Mem: 8GB
![image](https://github.com/ni/grpc-labview/assets/50438064/a0b8b5c4-0afa-4ce7-87df-617f9f2d6075)

### Steps to reproduce
1. Open attached project and then run the server/client example
![image](https://github.com/ni/grpc-labview/assets/50438064/7deef7ce-b108-4bd2-b0dc-493c2549b8c7)

2. When the test code is running, then swirch the windows focus to another app such as Task manager.
3. Then you will see the client example hang. Give focus back to LabVIEW will not fix this issue. The CPU usage of my virtual machine is ~50%, and never reach >70% during testing procedure.
![image](https://github.com/ni/grpc-labview/assets/50438064/afd0ce36-8fd9-4c4a-b88e-13a513fb0a50)

### Guess
This may be caused by CPU logic processor auto-assignment. My VMWare machine is not so powerful so it could reproduce this issue.

Any idea?

wendisong:
@yijueli, FYI

AndrewHeim:
I was experiencing something like this too, so I investigated.

**Situation:**
Very simple example using local loopback, wrap client in a For Loop for tens of thousands of iterations, reusing client grpc session between iterations.  Running in Windows 10 on a bare metal machine that isn't as resource constrained as your VM.   gRPC Library, Servicer, and Templates all version 1.0.1.1.

I'm seeing it hang occasionally, though not on the same iteration.  The client call appears to ignore the timeout indefinitely, but it is a hang - since closing the (apparently still responsive) server leaves the client unchanged, still hung.


**Possible solution:**
Race condition.  Here's the details:

[...]

PR #349, opened 2024-02-26
Fixed gRPC Client Hang issue

yash-ni:
Fix for #345 #330 #324 #320

Changes:

1. Fixed the race condition issue with occurrence hence solving a lot of client hang bugs
2. Added feature toggle for running the client without [LabVIEW Occurrence](https://www.ni.com/docs/en-US/bundle/labview-api-ref/page/menus/categories/programming/synchronization/occur-mnu.html).
3. Changed ActiveClientCalls from std::list to std::unordered_map hence resulting in a little performance increase.


src/grpc_client.h
ni-sujain, line 1: remove unnecessary auto format on function signatures

ni-sujain (approved):

Issue #366, opened 2024-06-26
LabVIEW crashes when attempting to unpack a repeated field.


ccaltagi:
GetUnpackedField.vim crashes when trying to unpack a repeated field.
Contact me for access to the code, Apparently, we cannot attach .vi or zip files.

AB#2786126

PR #369, opened 2024-07-18
Added support for unpacking varints in unknownfields.

yash-ni:
Fix for #366 

Changes:
* Implemented unpacking for all varint types (int32, int64, uint32, uint64, sint32, sint64, enum)


ni-sujain (approved):
you can mention that we don't support all types of unknown field parsing. This may fix the specific issue but cannot guarantee to work for all cases. If we want extensive support for unknown types, then it should be new feature request

Issue #361, opened 2024-05-30
LabVIEW Crashes when Stopping Two Parallel Running gRPC Servers (on Windows and on NI Linux RT)


gmkrtchy:
Hi,

 We are developing a solution as part of a TBU CCS project that incorporates ni gRPC. In simple terms, we have multiple modules that will have a server setup inside the module & we will be connecting clients from teststand to be able to communicate directly with these modules. These modules will be within a overriding application and running within NI LINUX RT.
 
We have come across an issue whereby multiple servers is causing NI RT and also Labivew on windows to crash. We do not believe we are doing anything out of the ordinary and we are doing something similar to the TBU SRD ReplayHiLAD RT application, of whcih we can see doesn't do anything "non-standard". 

As an example we created one Simple LV project consists of 2 Modules
 
![image](https://github.com/ni/grpc-labview/assets/106306776/db487a73-ed94-4201-a1c4-5ad8af841585)
![image](https://github.com/ni/grpc-labview/assets/106306776/e29cf226-207f-4518-9ac8-13e5493f13cb)

Both Servers has the same  code, after 30 seconds each server generates "Server Stop" event which stops the gRPC Loop.
I see that Module1 Server Stops properly, but during stopping Module 2 it loses the connection to the target (See screenshot below)

![image](https://github.com/ni/grpc-labview/assets/106306776/38d52986-95a4-4396-831b-dbd9af174b96)
![image](https://github.com/ni/grpc-labview/assets/106306776/fb552fc5-0a5c-46ca-8ece-2912214320c5)

We are using these versions of gRPC packages
![image](https://github.com/ni/grpc-labview/assets/106306776/9e6345b4-da39-4f78-8db5-ea21e03867f7)

There looks to be a similar ticket already created **[here](https://github.com/ni/grpc-labview/issues/356)** from a partner. Any help and advice on this topic is appreciated as this is a solution required for a key NI Customer
Please Find attached the LV Project with Protos
[gRPC_MultiServer.zip](https://github.com/ni/grpc-labview/files/15506004/gRPC_MultiServer.zip)

AB#2759169

nischalks:
@gmkrtchy, we will look into this issue.

j-medland:
This is a very rough guess and I am no C++ programmer but is it because `LVProtoServerReflectionPlugin` is being deleted here? 

Looking at how the plugin is instantiated and stored I believe it would be shared between both server instances.

[...]

Issue #356, opened 2024-03-22
Labview crashes when two (or more) servers are stopped


HouetIBS:
I'm using two servers and two clients in one application. When running the servers and clients, this works really well.

When I close one server (no matter which one I choose), the server is closed nicely as well.
But, when I close the second server, Labview crashes when calling the LVStopServer in the dll.
![image](https://github.com/ni/grpc-labview/assets/164482238/6c8280c7-b7a4-418f-a46e-4ff954ded35d)
(The gRPCid server is for both servers not 0 and unique for all servers)

If I just use the simple 'helloworld'-example and just copy an extra server in the application, it is really easy to reproduce the problem.
![image](https://github.com/ni/grpc-labview/assets/164482238/29f5ec85-2858-4c49-8fa6-70bd638479cc)

Would be great if this can be fixed.

Thanks in advance.

PS:
- Used LabVIEW 2021 32 and 64bit to verify this.
- Used gRPC-LabVIEW 1.2.0.1

With kind regards,

Erik

AB#2689710

nischalks:
Hi @HouetIBS, we wanted to check with you, which NI product team do you belong to ?

HouetIBS:
Hi,

I do not really understand your question. I work for IBS Precision Engineering in the Netherlands.
We use Labview for pretty much every application we build.

The current application, which uses gRPC, is written in Labview 2021 64bit.

I really like the gRPC/proto implementation to use as a communication standard between software products. Good that you all started a gRPC implementation for Labview.
Unfortunate, there are some challenges to overtake to make it a really good experience.

Met vriendelijke groeten,

Erik Houet
IBS Precision Engineering B.V.

nischalks:
Thats nice to know. Thanks Erik!

gmkrtchy:
Hi @nischalks,

For one of our Key Customer project we are using LV gRPC.
I faced the same issue, stopping two parallel running Servers caused LV Crash.
The same issue happened also on LabVIEW RT.
Any updates about the status of this issue? 
Thank you in advance.

Gor Mkrtchyan
Senior Solutions Engineer | CCS EMEA

SallyTYT:
Not sure if you are facing the same issue as this one https://github.com/ni/grpc-labview/issues/279

gmkrtchy:
@SallyTYT thanks for the comment,

[...]

PR #370, opened 2024-07-25
Remove singleton behavior for Reflection Plugin Class

yash-ni:
Fix for #356 and #361 

Changes:

1) Created a new singleton class "ProtoDescriptorString" that will save the current proto string descriptor and can be later used to initialize the reflection service.

2) Removed singleton behavior of Reflection plugin class "LVProtoServerReflectionPlugin" so that now every server will have its own reflection service and separate descriptor pools.


src/lv_interop.cc
```diff
@@ -44,6 +44,37 @@ namespace grpc_labview
 {
     grpc_labview::PointerManager<grpc_labview::gRPCid> gPointerManager;
 
+    //---------------------------------------------------------------------
+    //---------------------------------------------------------------------
+ 
+    // Initialize the static members
+    ProtoDescriptorString* ProtoDescriptorString::m_instance = nullptr;
+    std::string ProtoDescriptorString::descriptor_;
```
ni-sujain, line 52: how do we cleanup these things when server is stopped

yash-ni: Added functionality to free memory when all servers stop

src/lv_proto_server_reflection_plugin.h
```diff
@@ -39,6 +39,7 @@ namespace grpc_labview
 {
     class LVProtoServerReflectionPlugin : public ::grpc::ServerBuilderPlugin {
     public:
+        LVProtoServerReflectionPlugin(); // hide constructors TODO
```
ni-sujain, line 42: remove comment

ni-sujain (approved):

src/lv_proto_server_reflection_plugin.h
yash-ni: Done

PR #372, opened 2024-07-30
fixed bad alloc and added missing unix lv_interop function loading

j-medland:
I am working on some changes to the C++ build tooling and dll linkage (see #144) and noticed that the memory size allocation in `OccurServerEvent` function is likely 1 byte too big on 32-bit platforms and 3 bytes too small on 64-bit platforms (dependant on the LV array/string byte-packing) . You would have to be pretty unlucky to have something important allocated in the bytes that get "borrowed" on 64-bit systems but it seemed worth fixing.

As we are trying to create a LabVIEW-string from a `std::string` we can use the existing `SetLVString` function to allocate and copy the `std::string` into something we can use to call `PostUserEvent`. 

Clean-up of the handle is done with `DSDisposeHandle`

Strangley, `InitCallbacks` function only gets the address of `DSDisposeHandle` on Windows so I added the code so both platforms load the same functions.

Finally - I removed the call in `InitCallbacks` which would make the LabVIEW IDE or the RTE exit on a shared-library load error on Linux platforms. It is better to rely on LabVIEW's rudimentary shared-library call fault handling than to just make the whole IDE/Application exit with little indication to the user as to what happened but I am happy to discuss if you feel differently

Cheers


j-medland:
Sorry that the tests need to be rerun. I force pushed a commit with `--signoff` as per the contribution guide update

j-medland:
Hello - it would be great to get this reviewed or some more information if there is something I can do to help this happen?

I appreciate NI making this open source and I understand if your aren't in a position to accept community PRs but it would be great to know. I would also be happy if this was just fixed by one of the maintainers.

jasonmreding (approved):

PR #377, opened 2024-08-26
Avoid proxy issues for localhost only communication

jasonmreding:
# Description

This PR attempts to avoid grpc connection issues due to http proxies when communicating over loopback addresses. If we detect the target address for a gRPC channel is the localhost or equivalent, then we explicitly disable sending traffic through the proxy.

# Implementation
- Added a git submuodule for [asio](https://github.com/chriskohlhoff/asio) which is what boost is based on. I attempted to use boost directly, but I couldn't figure out how to get headers only build to work in a reasonable way so I fell back to this solution which I found to be simpler.
- Updated cmake files to add new include directories for the headers which are being used from `asio`
- Updated `LabVIEWgRPCClient::Connect` to set `GRPC_ARG_ENABLE_HTTP_PROXY` to false if the target address is detected to be a loopback address.

# Testing
I didn't see much in the way of an automated regression test framework so I didn't create any automated tests as part of this change. I manually tested that the following addresses worked and used the expected proxy settings:
- localhost
- localhost:42000
- 127.0.0.1
- 127.0.0.1:42000
- \[::1]
- \[::1]:42000
- http://localhost:42000
- foo


src/grpc_client.cc
```diff
@@ -23,6 +25,48 @@ namespace grpc_labview
...
+        }
+
+        std::string localhost("localhost");
+        bool isLoopback = hostname.size() == localhost.size() && std::equal(hostname.begin(), hostname.end(), localhost.begin(), ichar_equals);
+        if (!isLoopback)
+        {
+            try
+            {
+                asio::ip::address ip_address = asio::ip::make_address(hostname);
+                isLoopback = ip_address.is_loopback();
```
bkeryan, line 60: Do you really need boost.asio for this?

You can use [inet_pton](https://www.man7.org/linux/man-pages/man3/inet_pton.3.html) to try parsing the address as both IPv4 and IPv6. 
- For IPv4, use `AF_INET` and `struct in_addr`. To check for loopback, I think you can use  something like `(ntohl(x.s_addr) & 0xFF000000) == (INADDR_LOOPBACK & 0xFF000000)` (untested).
- For IPv6, use `AF_INET6` and `struct in6_addr`. To check for loopback, use the [IN6_IS_ADDR_LOOPBACK](https://man7.org/linux/man-pages/man0/netinet_in.h.0p.html) macro.

src/grpc_client.cc
```diff
@@ -23,6 +25,48 @@ namespace grpc_labview
...
+        {
+            size_t lastColonPosition = hostname.rfind(':');
+            if (lastColonPosition != std::string::npos)
+            {
+                hostname = hostname.substr(0, lastColonPosition);
+            }
+        }
+
+        std::string localhost("localhost");
+        bool isLoopback = hostname.size() == localhost.size() && std::equal(hostname.begin(), hostname.end(), localhost.begin(), ichar_equals);
```
bkeryan, line 54: This is a roundabout way to compare strings. If you're going to use the current locale for case folding, you might as well use [strcasecmp](https://www.man7.org/linux/man-pages/man3/strcasecmp.3.html). Calling `tolower` on each letter may do the wrong thing in some cases, such as [German ß](https://en.wikipedia.org/wiki/%C3%9F)).

j-medland:
Could the user just specify a local variable to specify if the proxy should be ignored for localhost etc?

https://github.com/grpc/grpc/blob/591d56e1300b6d11948e1b821efac785a295989c/doc/environment_variables.md

jasonmreding:
> Could the user just specify a local env variable to specify if the proxy should be ignored for localhost etc?
> 
> https://github.com/grpc/grpc/blob/591d56e1300b6d11948e1b821efac785a295989c/doc/environment_variables.md

Yes, that is an option and is what we have been telling customers to avoid these issues. However, the errors you get when this is a problem aren't very helpful, and a lot of time is wasted via support channels before getting to the bottom of what is wrong.

src/grpc_client.cc
jasonmreding: The underlying headers/source ends up using that library so yes I'm sure it can be done. I was trying to avoid reinventing the wheel and learning about the minutiae of that lower level library. Since this is headers only and doesn't require external linkage, I didn't really seem the harm to using it.

BTW, this isn't technically boost.asio. It's just asio. Some of the differences are detailed [here](https://think-async.com/Asio/AsioAndBoostAsio.html). My understanding is that boost.asio is occasionally updated from asio.

src/grpc_client.cc
jasonmreding: I'm rusty on C++ and maybe I'm being dense, but `strcasecmp` isn't part of the C++ standard library is it. At least when I tried to use it, I got errors saying "error C1083: Cannot open include file: 'strings.h': No such file or
 directory". I don't care about locales in this case since I only care about matching against this one string. Maybe trying to ignore case here is going overboard?

src/grpc_client.cc
bkeryan: Okay, plain asio sounds lighter-weight than boost.asio, so I'll close this.

BTW, I was curious whether gRPC has a function for this, and it does, but it's an internal implementation detail and it only checks against 127.0.0.1 / ::1, not other addresses in the loopback subnets.
https://github.com/grpc/grpc/blob/f34b4faca4bbad70ddee456b6e2cfd156a253392/src/core/server/xds_server_config_fetcher.cc#L955

src/grpc_client.cc
bkeryan: Oops, `strcasecmp` is a BSD function. The Microsoft equivalent is `_stricmp`. If you search NI's codesearch for these, you will find a lot of #ifdefs.

> I don't care about locales in this case since I only care about matching against this one string. Maybe trying to ignore case here is going overboard?

Comparing to localhost without using the name resolver is kind of a hack anyway. 

It would be easier to ignore SpOnGeBoB cAsE and do `hostname == "localhost" || hostname == "LOCALHOST"` :)

bkeryan (approved):

src/grpc_client.cc
jasonmreding: Fair enough. My goal is to make this "just work" for what I would expect to be the majority of cases where loopback communication is used. If use of DNS host names for the localhost proves to be a support issue, we can revisit and improve this further. I'd rather not add the overhead of a DNS lookup to every connection at this point.

jasonmreding:
@yash-ni @ni-sujain It looks like you have the most recently submitted PRs to the repo. Can you help me find someone to review this PR and/or give me write access to the repo so I can see who approving reviewers are?

ni-sujain:
> @yash-ni @ni-sujain It looks like you have the most recently submitted PRs to the repo. Can you help me find someone to review this PR and/or give me write access to the repo so I can see who approving reviewers are?

I have given you access. Thanks

Issue #378, opened 2024-08-29
Nested clusters get jumbled from service to client


BrandtSwanson:
There is an issue when populating a response with repeated messages. array_0 should have 1,2,3,4 and array_1 should have 5,6,7,8, but all the data ends up in array 0.

Here is the proto file:
![Proto](https://github.com/user-attachments/assets/3918dae6-a7e2-43d0-96f8-2c958fa38b79)

Here you can see how the response is populated:
![GetStatus](https://github.com/user-attachments/assets/47b59380-26a1-45b0-acbd-e613df094a30)

Here is how the response shows up on the client side:
![Response](https://github.com/user-attachments/assets/47971d05-6d85-430f-9472-23ebfd65b4ba)

AB#2842884

jasonmreding:
Setting `EfficientMessageCopy = FALSE` in `feature_config.ini` will avoid this problem until bugs in the "efficient" code path can be fixed.

jasonmreding:
This should be fixed in #387

Issue #384, opened 2024-09-23
LabVIEW crashes when serializing nested messages


jasonmreding:
See https://github.com/ni/measurement-plugin-labview/issues/594 for a description of the crash and how to reproduce. The proto API for the crash is [here](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/discovery/v1/discovery_service.proto#L39).

After doing a bit of debugging and trial and error, the crash only happens if the service reports results where the [annotations](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/discovery/v1/discovery_service.proto#L96) field is populated in the response. Also, setting `EfficientMessageCopy = FALSE` in the ini file also fixes the crash. The crash is being triggered from the following DAbort in LV:

```C++
TH_REENTRANT size_t _FUNCC DSGetHandleSize(const void* h)
{
    size_t size;

    if (!GoodDSHandleOpt(h)) {
        gDAbortID(0xF50EFD7BUL);
        return 0;
    }
    size = BGetHandleSize(gMemZone, h);
    return size;
}
```

This seems to be the result of #335 and #352. I suspect #378 is also related to this.

AB#2864684

jasonmreding:
@pratheekshasn @ni-sujain @CPattar-NI @yash-ni 

I have been debugging this issue a bit, and I'm having a hard time understanding how the code works or what this change was intended to provide over the original implementation. I guess it was purely a performance improvement? From what I can see, this code path doesn't appear to be assigning data to the cluster from memory allocated by LV. Instead, it looks like it is copying data from the `LVMessage` itself. If that's the case, then it's not surprising we are running into a crash, but maybe I just don't understand the code. It also appears the `NumericArrayResize` is allocating 8x the amount of memory necessary for a repeated nested message. However, that is happening in both code paths (with and without the toggle enabled) so maybe I'm misreading the code. While that is wasteful, it is not inherently incorrect from a functional standpoint. However, it might be indicative of other problems lingering in the code and code that isn't quite ready for release.

Given the number of issues already caused by this change and the complexity of the code, I'm wondering if we should just disable the feature or if you have any other insights/suggestions for how to fix the current, active code path? I guess this change was made as a performance optimization? If so, I don't have any background on how crucial the optimization was or what impact rolling it back might have on clients.

jasonmreding:
This should be fixed in #387

PR #387, opened 2024-09-30
Several bug fixes for serialization or nested/repeated messages

jasonmreding:
# Description

This PR fixes several issues with message serialization when `EfficientMessageCopy = TRUE`.
#378 
#384
#392 

# Implementation
- Fixed several places where the array being allocated was larger than required. In effect, we were requesting an allocation for the total byte size when `NumericArrayResize` takes the element size and not the byte size.
- Fixed a few places where arrays were always allocating 8 byte pointers rather than 4 bytes for a pointer on 32 bit platforms.
- Cleaned up some state management for better consistency with how we handle repeated string and message fields.
- Fixed issues where the buffer used to copy data for repeated nested messages was always copying previous elements in the array and copying the address for the same buffer multiple times in the message structure. This leads to DWARNs in the memory manager since we attempt to deallocate the same block of memory multiple times. I essentially fixed this by allocating a new message on the stack for each element of the repeated field rather than reusing the same message object for all elements of the repeated field. I ultimately found this to be easier to read and rationalize about even though it probably requires more memory allocations. The alternative would be to clear the maps / repeated fields we use to hold the data as we parse the data. This would allow us to reuse those objects at their existing capacity which more than likely enough to hold the next element as well. However, the readability of the code is worse. If anyone feels like that performance gain is worth it, I can look into making that change.
- Fixed issues where the tag size calculation for repeated string/message fields was incorrect and would only work if the tag could be encoded in a single byte.

# Testing
- I manually tested the proto file attached to #378 failed before the changes and works as expected afterwards. 
- I verified the proto API referenced from #384 works as expected and no longer crashes using either 32 or 64 bit LV on Windows.
- I validated messages with enough fields as described by #392 pack/unpack correctly.
- I created a test proto with lots of nested repeated string and message fields in the response message. I then generated fairly large arrays for those fields in the server. Using the debugger, I verified that the repeated fields for both strings and messages were broken up (not sent in a contiguous block), and that the new code handles that appropriately.


src/cluster_copier.cc
```diff
@@ -287,7 +287,7 @@ namespace grpc_labview {
             auto repeatedString = static_cast<const LVRepeatedMessageValue<std::string>&>(*value);
             if (repeatedString._value.size() != 0)
             {
-                NumericArrayResize(0x08, 1, start, repeatedString._value.size());
+                NumericArrayResize(GetTypeCodeForSize(sizeof(char*)), 1, start, repeatedString._value.size());
```
jasonmreding, line 290: Note: An 8 byte pointer is too large for 32 bit LV.

src/cluster_copier.cc
```diff
@@ -330,14 +330,21 @@ namespace grpc_labview {
...
+                auto byteSize = repeatedNested->_value.size() * clusterSize;
+                auto alignment = nestedMetadata->alignmentRequirement;
+                auto alignedElementSize = byteSize / alignment;
+                if (byteSize % alignment != 0)
+                {
+                    alignedElementSize++;
+                }
 
-                NumericArrayResize(0x08, 1, start, repeatedNested->_value.size() * clusterSize);
+                NumericArrayResize(GetTypeCodeForSize(alignment), 1, start, alignedElementSize);
```
jasonmreding, line 341: Note: The previous allocation was 8x larger than it needed to be since the element size passed was the byte size, and we were specifying an element size of 8 bytes. We now pick an element size that meets our alignment requirements and update the number of elements to allocate accordingly.

src/lv_message_efficient.cc
```diff
@@ -67,25 +67,17 @@ namespace grpc_labview
...
-         
-            // Get the _repeatedMessageValues vector from the map
-            auto _repeatedStringValuesIt = _repeatedStringValuesMap.find(key);
-            if (_repeatedStringValuesIt == _repeatedStringValuesMap.end())
-            {
-                _repeatedStringValuesIt = _repeatedStringValuesMap.emplace(key, google::protobuf::RepeatedField<std::string>()).first;
-            }
-
+        {
+            auto repeatedString = google::protobuf::RepeatedField<std::string>();
```
jasonmreding, line 76: NOTE: There is no reason to hold onto this repeated string field beyond this stack frame. I've removed the backing fields from the message that were being used to track it and just allocate everything on the stack now for efficiency/clarity.

src/lv_message_efficient.cc
```diff
@@ -94,20 +86,21 @@ namespace grpc_labview
             } while (ExpectTag(tag, protobuf_ptr));
 
 
-            NumericArrayResize(0x08, 1, reinterpret_cast<void*>(const_cast<char*>(lv_ptr)), _repeatedStringValuesIt->second.size());
+            NumericArrayResize(GetTypeCodeForSize(sizeof(char*)), 1, reinterpret_cast<void*>(const_cast<char*>(lv_ptr)), repeatedString.size());
```
jasonmreding, line 89: NOTE: A pointer on 32 bit platforms is not 8 bytes.

src/lv_message_efficient.cc
```diff
@@ -117,82 +110,78 @@ namespace grpc_labview
...
             auto numElements = 128;
             auto elementIndex = 0;
             auto clusterSize = metadata->clusterSize;
-            auto alignment = metadata->alignmentRequirement;
             auto arraySize = numElements * clusterSize;
             char _fillData = '\0';
 
             // Get the _repeatedMessageValues vector from the map
-            auto _repeatedMessageValuesIt = _repeatedMessageValuesMap.find(metadata->messageName);
+            auto _repeatedMessageValuesIt = _repeatedMessageValuesMap.find(fieldInfo.fieldName);
```
jasonmreding, line 136: NOTE: This effectively fixes #378 where a message has two repeated fields of the same message type. Since we were tracking the data via the message name rather than the field name, the second field was adding its data to the end of the first field.

src/lv_message_efficient.cc
```diff
@@ -117,82 +110,78 @@ namespace grpc_labview
...
             {
                 auto m_val = std::make_shared<RepeatedMessageValue>(fieldInfo, google::protobuf::RepeatedField<char>());
-                _repeatedMessageValuesIt = _repeatedMessageValuesMap.emplace(metadata->messageName, m_val).first;
+                _repeatedMessageValuesIt = _repeatedMessageValuesMap.emplace(fieldInfo.fieldName, m_val).first;
+                _repeatedMessageValuesIt->second.get()->_buffer.Resize(arraySize, _fillData);
             }
+            else
+            {
+                // Zero out data from the previous use of the buffer.
+                auto bytesUsed = _repeatedMessageValuesIt->second.get()->_numElements * clusterSize;
```
jasonmreding, line 146: NOTE: Rather than using `_continueFromIndex` to begin writing from where the previous element in the array left off, we now always reset the index back to zero. For that to work, we need to explicitly zero out any data from the previous element. Otherwise, if some fields weren't serialized because they were set to their default values, we can end up reusing the data from the previous element which leads to data corruption. I thought about just allocating a new buffer each time to simplify things, but I decided the performance gains of not repeatedly allocation a buffer were probably worth it.

The problem with the previous approach is that while we were effectively tracking the end index of the buffer, we weren't also tracking the beginning index. So when we finally copy the data in `PostInteralParseAction`, we always copy from index zero which includes data from all of the elements that preceded it. Also, the previous approach allocated ever growing memory rather than reuse the memory from previous allocations once it had been copied.

src/lv_message_efficient.cc
```diff
@@ -204,27 +193,38 @@ namespace grpc_labview
     {
         for (auto nestedMessage : _repeatedMessageValuesMap)
         {
+            if (!nestedMessage.second.get()->_validData)
```
jasonmreding, line 196: NOTE: This check ensures we only copy the data into a LV data structure once. In particular, it handles cases where the message element at index zero in the array had a non-default value for a repeated message field but the message element at index one had an empty array. In that case, we don't want to copy the array from the previous element again.

src/lv_message_efficient.cc
```diff
@@ -204,27 +193,38 @@ namespace grpc_labview
...
 
             auto metadata = fieldInfo._owner->FindMetadata(fieldInfo.embeddedMessageName);
             const char* lv_ptr = (this->GetLVClusterHandleSharedPtr()) + fieldInfo.clusterOffset;
             auto clusterSize = metadata->clusterSize;
             auto alignment = metadata->alignmentRequirement;
 
 
-            // shrink the array to the correct size
-            auto arraySize = numElements * clusterSize;
-            NumericArrayResize(0x08, 1, reinterpret_cast<void*>(const_cast<char*>(lv_ptr)), arraySize);
```
jasonmreding, line 219: NOTE: Again, this allocation is 8x larger than it needs to be.

src/lv_message_efficient.h
```diff
@@ -86,13 +84,14 @@ namespace grpc_labview
             // copy into LVCluster
             if (numElements != 0)
             {
-                NumericArrayResize(0x08, 1, reinterpret_cast<void*>(const_cast<char*>(_lv_ptr)), numElements);
```
jasonmreding, line 89: NOTE: This was over allocating for any element type that wasn't 8 bytes (so basically anything other than doubles).

src/proto_parser.cc
```diff
@@ -262,18 +262,19 @@ LIBRARY_EXPORT int LVGetServices(grpc_labview::LVProtoParser* parser, grpc_labvi
     }
 
     auto count = parser->m_FileDescriptor->service_count();
-    if (NumericArrayResize(0x08, 1, services, count * sizeof(ServiceDescriptor*)) != 0)
+    auto elementSize = sizeof(ServiceDescriptor*);
+    if (NumericArrayResize(grpc_labview::GetTypeCodeForSize(elementSize), 1, services, count * elementSize) != 0)
```
jasonmreding, line 266: NOTE: All of the changes to this file are just white space changes other than the five calls to `NumericArrayResize` in this file. They now all calculate the pointer size rather than always allocating 8 bytes for a pointer. I'm less certain about these changes since I'm a little confused as to why we are storing pointers to non LV data structures via array handles allocated by the LV memory manager. I'm also not certain who is calling these exported functions. The APIs appear to be related to grpc server reflection, but I don't know where/how they are being called.

dixonjoel:
I noticed these warnings in the PR build. Probably pre-dates your changes, but I wonder if they're indicating problems or can be ignored.
```
D:\a\grpc-labview\grpc-labview\third_party\grpc\third_party\upb\upb/table_internal.h(200,28): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) [D:\a\grpc-labview\grpc-labview\build\grpc\upb.vcxproj]
D:\a\grpc-labview\grpc-labview\third_party\grpc\third_party\upb\upb/msg_internal.h(469,1): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) [D:\a\grpc-labview\grpc-labview\build\grpc\upb.vcxproj]
D:\a\grpc-labview\grpc-labview\third_party\grpc\third_party\upb\upb/msg_internal.h(741,34): warning C4090: 'function': different 'const' qualifiers [D:\a\grpc-labview\grpc-labview\build\grpc\upb.vcxproj]
D:\a\grpc-labview\grpc-labview\third_party\grpc\third_party\upb\upb\decode.c(190,61): warning C4646: function declared with 'noreturn' has non-void return type [D:\a\grpc-labview\grpc-labview\build\grpc\upb.vcxproj]
D:\a\grpc-labview\grpc-labview\third_party\grpc\third_party\upb\upb\decode.c(412,31): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) [D:\a\grpc-labview\grpc-labview\build\grpc\upb.vcxproj]
  gzclose.c
  absl_random_internal_platform.vcxproj -> D:\a\grpc-labview\grpc-labview\build\grpc\third_party\abseil-cpp\absl\random\Release\absl_random_internal_platform.lib
  def.c
```

dixonjoel (approved):

dixonjoel:
@bkeryan @jasonmreding I don't think we can afford to wait much longer for a review. Can Brad just complete his review and then we can just move forward?

src/lv_message_efficient.cc
CPattar-NI: The reason we had to maintain the lifetime of the repeated field longer than the stack was for much larger messages. When we do have very large messages, gRPC breaks the data into smaller chunks in memory, and parsing the message happens with multiple calls to `LVMessage::_InternalParse`, and not on a single call.
Could you try running tests with large repeated messages?

jasonmreding: I'm not sure I follow. What would be considered a large repeated message for this case? I assume you are talking about any message with a repeated string field of sufficiently large length? I think the wire protocol technically allows for the same field to be sent multiple times where last one wins for scalar fields and values are appended for repeated fields. Is that what you are referring to? If so, I might have to think on that some more. I'm not really sure how to test that without writing the byte stream directly.

jasonmreding: I think I see what you are talking about. I guess in the case of repeated strings, we were potentially doing multiple copies into the LV array rather than one copy from the repeated field into the LV array at the very end. I find reusing the message on the stack makes it really difficult to reason about.

enc_temp_folder/ecfc2145d7dbe55dcc4fe35c069b538/lv_message_efficient.cc
dixonjoel, line 1: Is this file from the build or something? This looks strange.

jasonmreding: I have no idea what generated that. It should be gone now.

src/lv_message.cc
```diff
@@ -356,10 +356,12 @@ namespace grpc_labview
             {
                 v = std::static_pointer_cast<LVRepeatedMessageValue<std::string>>((*it).second);
             }
-            protobuf_ptr -= 1;
+
+            auto tagSize = CalculateTagWireSize(tag);
```
jasonmreding, line 360: NOTE: This no longer assumes the field can be encoded in a one byte tag. This change was made for both repeated string and message fields for both the original serialization code path and the efficient message code path.

src/lv_message_efficient.cc
jasonmreding: The latest changes revert back to something closer to the original. I've changed the implementation so repeated strings and messages more or less work the same so hopefully the code is easier to read and reason about. For this to work in a more straight forward way, I now allocate a new nested message on the stack for each element in the repeated field rather than reusing it and dealing with state from the previous element polluting the data for the next element.

src/lv_message_efficient.cc
```diff
@@ -204,27 +185,52 @@ namespace grpc_labview
...
         }
-    }
 
+        for (auto repeatedStringValue : _repeatedStringValuesMap)
+        {
+            auto& fieldInfo = repeatedStringValue.second.get()->_fieldInfo;
+            auto& repeatedString = repeatedStringValue.second.get()->_repeatedString;
+            const char* lv_ptr = (this->GetLVClusterHandleSharedPtr()) + fieldInfo.clusterOffset;
+
+            NumericArrayResize(GetTypeCodeForSize(sizeof(char*)), 1, reinterpret_cast<void*>(const_cast<char*>(lv_ptr)), repeatedString.size());
```
dixonjoel, line 221: Could use 'numDimensions' here instead of '1' if you keep the variable.

dixonjoel (approved):

src/lv_message_efficient.cc
```diff
@@ -204,27 +185,52 @@ namespace grpc_labview
...
-            auto arraySize = numElements * clusterSize;
-            NumericArrayResize(0x08, 1, reinterpret_cast<void*>(const_cast<char*>(lv_ptr)), arraySize);
+            // Allocate an array with the correct size and alignment for the cluster.
+            auto byteSize = numClusters * clusterSize;
+            auto alignedElementSize = byteSize / alignment;
+            if (byteSize % alignment != 0)
+            {
+                alignedElementSize++;
+            }
+            auto numDimensions = 1;
```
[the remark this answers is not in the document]
bkeryan: C++ doesn't have named parameters. Most code uses some variation on `/* numDims */ 1` (where `numDims` is the actual parameter name). Given the number of times it's used, I don't mind hardcoding 1 in the argument list without a comment.

bkeryan (changes requested):

src/cluster_copier.cc
```diff
@@ -330,14 +330,21 @@ namespace grpc_labview {
...
                 auto array = *(LV1DArrayHandle*)start;
                 (*array)->cnt = repeatedNested->_value.size();
```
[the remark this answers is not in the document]
jasonmreding: Your understanding is more or less correct. I wouldn't consider myself to be that knowledgeable on the nuances between using `NumericArrayResize` vs. `DSSetHandleSize`. I suspect you might need to use [DSNewHandle](https://www.ni.com/docs/en-US/bundle/labview-api-ref/page/properties-and-methods/lv-manager/dssethandlesize.html?srsltid=AfmBOooRebmUqvdTjIppBAtJYKkAYS4w1Gzm_hVDj9I6GNgTaZkop3cf) as I'm not sure if `DSSetHandleSize` can only resize a valid handle vs. also allocating a new handle. It's also unclear how much alignment logic is present in those functions vs. `NumericArrayResize`. The documentation for `NumericArrayResize` states:

> Resizes a data handle that refers to a numeric array. This routine also accounts for alignment issues. It does not set the array dimension field. If *dataHP is NULL, LabVIEW allocates a new array handle in *dataHP.

The documentation for the other memory manager functions does no't explicitly state this so I'm not sure how much the methods will do for you and how much you have to figure out for yourself. #352 included changes which explicitly changed from `DSSetNewHandle` to `NumericArrayResize`, but I'm not entirely sure why. I had enough other things to worry about that I didn't bother poking into this further.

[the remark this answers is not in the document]
bkeryan: > I suspect you might need to use [DSNewHandle](https://www.ni.com/docs/en-US/bundle/labview-api-ref/page/properties-and-methods/lv-manager/dssethandlesize.html?srsltid=AfmBOooRebmUqvdTjIppBAtJYKkAYS4w1Gzm_hVDj9I6GNgTaZkop3cf) as I'm not sure if DSSetHandleSize can only resize a valid handle vs. also allocating a new handle.

That's what I meant by "It takes a handle rather than an hp..." If you want the same behavior as NumericArrayResize where you can start with a null handle, you have to write something like `if (h) DSSetHandleSize(h, ...) else h = DSNewHandle(...)`. 

> It's also unclear how much alignment logic is present in those functions vs. NumericArrayResize.

[DSNewHandle](https://www.ni.com/docs/en-US/bundle/labview-api-ref/page/properties-and-methods/lv-manager/dsnewhandle.html) says

> The routine aligns all handles and pointers in DS to accommodate the largest possible data representations for the platform in use.

I checked the source and this is at least 8.

Anyway, I don't see any problems with this use of NumericArrayResize.

src/lv_message.cc
```diff
@@ -400,6 +402,19 @@ namespace grpc_labview
         }
     }
 
+    //---------------------------------------------------------------------
+    //---------------------------------------------------------------------
+	int LVMessage::CalculateTagWireSize(google::protobuf::uint32 tag)
+	{
+        // The current gRPC libraries only support a maximum number of field numbers that
```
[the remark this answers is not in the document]
jasonmreding: I guess it was based on this implementation of [ExpectTag](https://github.com/protocolbuffers/protobuf/blob/6345cbed1acb33506c12f592fa4d6f27890db4b4/src/google/protobuf/parse_context.h#L615) which is code that is [no longer present](https://github.com/protocolbuffers/protobuf/commit/6d03f9b19bd7aa10d541f6c1a11ac71f2851eece) in the mainline. It was also based on the implementation of [ExpectTag](https://github.com/ni/grpc-labview/blob/master/src/lv_message.cc#L390) within our own repo.  

However, I see now that this probably a limitation of our own making. I will update it to use this [implementation](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/io/coded_stream.h#L843) instead. However, if there are no objections, I will hold off trying to fix our implementation of `ExpectTag` as part of this change.

src/lv_message_efficient.cc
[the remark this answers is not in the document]
jasonmreding: I reverted back to just using a constant for consistency with the rest of the code.

src/lv_message_efficient.cc
jasonmreding: Sticking with the constant.

bkeryan (approved):

src/lv_message.cc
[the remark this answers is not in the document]
bkeryan: Okay, that works too.

Issue #396, opened 2024-11-19
oneof fields are not read from the client correctly


jasonmreding:
Using the following proto file:

```protobuf
syntax = "proto3";

package oneof;

service EchoOneofService {
  rpc EchoOneof(EchoOneofRequest) returns(EchoOneofResponse);
}

message AttributeValue
{
  oneof attribute
  {
    bool bool_value = 1;
    int32 integer_value = 2;
    double double_value = 3;
    string string_value = 4;
  }
}

message EchoOneofRequest {
 AttributeValue attribute = 1;
}

message EchoOneofResponse {
 AttributeValue attribute = 1;
}
```

generate a client and a server. Modify the server to just pass back the same data as it received from the client. Create a client that sets one of the fields comprising the oneof and then invoke the `EchoOneof` RPC. If you step into the generated LV code around the RPC and probe the class wire, you'll see the index and data are updated appropriately for the class in the request data. However, the index field for the class in the response isn't set even though the one of field has the appropriate data value. The flat->rich conversion then reassigns a default class value since the index is 0 rather than the appropriate index for the field number. I suspect the C++ client code is failing to assign the index based on the last set field value on the wire. This might also be related to #343?

AB#2936036

nischalks:
@jasonmreding let us know incase you have a fix, we can group a bunch of bugs together and make a release for you. Thanks!

jasonmreding:
This should be resolved in https://github.com/ni/grpc-labview/releases/tag/v1.3.0.1

PR #397, opened 2024-11-19
Cleaning front panel and block diagrams to improve readability. Where…

jasonmreding:
# Description

I am planning to make changes to the code generators in up coming PRs. Before doing so, I am doing a mass cleanup of the existing VIs in this PR to improve readability and minimize diffs for more meaningful changes in the future.

# Implementation
There are no intended functional changes as part of this PR. The changes are purely cosmetic in nature to improve readability. Move control/indicator label to the left/right on the block diagram, uncross wires, and that sort of thing. In some cases, common code was extracted and unreferenced code was deleted. There were a few VIs that didn't have source only setting and those have been fixed as well. I also updated the icon for the error code VIs as a solid red background color doesn't play well with the icon editor algorithm in later LV versions. It tends to display the text over the library banner rather than limiting the text to the area below the banner.

# Testing
- Manually tested round trip of data between client/server for unary API using repeated message fields.
- Manually tested round trip of data between client/server for unary API using oneof field. Filed #396 as a result of this testing. Verified the same behavior is present before these changes. The generated code looks good though.
- Manually tested code generation of [measurement_service.proto](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/measurement/v2/measurement_service.proto) works for both client and server.


labview source/Client Server Support New/gRPC Scripting Tools/Error Handling/Error 5000.vi
dixonjoel, line 1: nit: Ctrl + U on the front panel would make it possible to see the .vi name when opening these...
image

dixonjoel (approved):
In general, Ctrl + U on a lot of the front panel would be nice. But that's very minor. Feel free to submit without any additional cleanup.

PR #399, opened 2024-12-04
Cleaning up generated LV code and restricting scope as warranted

jasonmreding:
# Description

Cleaning up generated LV code for readability and restricting access scope where appropriate. I also modified the generated code for the server such that the concrete service instances are created by default and will handle multiple service APIs without requiring manual coding from the user.

# Implementation
- Changes to template VIs (client, server, and oneof) were just standard cleanup of block diagrams and front panel. Access scope for some of the templates were changed to private/protected as warranted.
- Inserted BD/FP cleanup calls to the scripted code any time new nodes were scripted or typedef constants/controls were updated.
- Added scripting calls to change access scope of generated VIs / typedefs as warranted. In general, all of the "flat" typdef and serialization VIs are marked as private implementation details.
- Updated server template so that the generator automatically scripts creation of each concrete service class. The template then handles spinning up the service classes in parallel and synchronizes shutdown across all instances. This was mostly stolen from what is done in the [measurement-plugin-labview](https://github.com/ni/measurement-plugin-labview) repo with measurement services.

# Testing
Manually tested generation of both the client and the server using proto APIs that included nested messages, enums, and oneof fields.


labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Get Or Create Typedef.vi
dixonjoel, line 1: nit: It would be nice if the top error wire would be cleaned up to be straight through.
image

dixonjoel:
One general concern about cleaning up the front panel and block diagram. Sometimes it can change the diagram quite a bit, e.g. from the way we arranged it in the client or server template. If we spend effort making the template VIs nice (which doesn't match what cleanup diagram would do), we may be wasting effort there.

So, are you running cleanup on every template VI we instantiate? Or only ones that you modify significantly. Have you compared the cleaned up VIs with the template VIs?

I don't think this is a major concern, just wanted to bring it up for discussion.

dixonjoel (approved):
Approved with minor suggestions / discussion.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Get Or Create Typedef.vi
jasonmreding: I have gone back and forth on this, and there really isn't a great style IMO to make scripting code in LV readable. The two general approaches I've seen are:
1. Keep the error wires straight.
2. Keep the refnum wires straight.

With the first approach, you end up having to "leap frog" refnum wires over nodes in order to keep the error wire straight without wires running behind nodes. This is OKish when you are only dealing with a couple of refnum wires/types. However, once you get 3 or 4 different refnums, I find the code really difficult to read since all the wires are the same color and is difficult to maintain track of which wire is which. So unless you use the help window and keep hovering over wires to remind yourself which wire is which you would be completely lost.

With the second approach, you end up with straight horizontal refnum wires such that it is easier to track/remember which wire is which. However, you end up with error wires that kind of zigzag up and down the diagram. In some cases, you can "pretty this" up a bit by running parallel error wires with `Merge Errors` at the end. However, this is sometimes more difficult to think about from an error handling perspective in terms of cleanup.

So while I'm not always consistent on which of the two approaches above that I use, I've been leaning more towards the second approach lately, especially for more complicated diagrams like this one where there are 3+ refnums at play at the same time that are otherwise indistinguishable by just looking at the wire. I realize there is a lot of subjectivity involved here.

jasonmreding:
> One general concern about cleaning up the front panel and block diagram. Sometimes it can change the diagram quite a bit, e.g. from the way we arranged it in the client or server template. If we spend effort making the template VIs nice (which doesn't match what cleanup diagram would do), we may be wasting effort there.
> 
> So, are you running cleanup on every template VI we instantiate? Or only ones that you modify significantly. Have you compared the cleaned up VIs with the template VIs?
> 
> I don't think this is a major concern, just wanted to bring it up for discussion.

My thoughts on this:
- You can never really go wrong with applying FP cleanup.
- Even if we run BD cleanup on the template VI, starting with a clean layout of the template is still helpful from a readability/maintainability standpoint.
- We don't do BD cleanup on all template VIs, only the ones where we script new nodes on the diagram or replace place holder constants/controls with the generated typedef reference. Most of these templates are simple enough that BD cleanup gives you something close or good enough to what you might layout by hand. It is certainly better than what you get with scripting nodes on top of each other. Calculating positions so nodes don't overlap without using BD cleanup is otherwise difficult/tedious. If you run into a situation where you have a significant portion of a diagram that is "static" and you want to manually lay it out in addition to a significant amount of scripted code you want to "auto layout", then I think the recommended practice is to extract all of the scripted code into a sub VI and use the VI boundary to isolate the cleanup actions.
- There are probably more opportunities for further cleanup. For instance, there are cases where we update constants like string constants and don't re-layout because of that. This can lead to overlap/crowding on the block diagram. Similarly, there are cases where cluster constants on the block diagram only serve as a type specifier. In those cases, we could save space and the BD cleanup would be more effective by using the icon view for cluster. However, I think this setting gets lost when doing replacements, and I haven't taken the time to try and enforce it with scripting yet. Maybe in a future PR.

Anyway, I've audited generated code for both the client and the server, and I think these changes are an improvement in almost all cases. If you look at before and after comparisons of the generated code with this change, you can see for yourself. I couldn't think of a great way to easily show that with pictures, but I can attach some after pictures if there is something in particular you wanted to see.

dixonjoel:
> > One general concern about cleaning up the front panel and block diagram. Sometimes it can change the diagram quite a bit, e.g. from the way we arranged it in the client or server template. If we spend effort making the template VIs nice (which doesn't match what cleanup diagram would do), we may be wasting effort there.
> > So, are you running cleanup on every template VI we instantiate? Or only ones that you modify significantly. Have you compared the cleaned up VIs with the template VIs?
> > I don't think this is a major concern, just wanted to bring it up for discussion.
> 
> My thoughts on this:
> 
> * You can never really go wrong with applying FP cleanup.
> * Even if we run BD cleanup on the template VI, starting with a clean layout of the template is still helpful from a readability/maintainability standpoint.
> * We don't do BD cleanup on all template VIs, only the ones where we script new nodes on the diagram or replace place holder constants/controls with the generated typedef reference. Most of these templates are simple enough that BD cleanup gives you something close or good enough to what you might layout by hand. It is certainly better than what you get with scripting nodes on top of each other. Calculating positions so nodes don't overlap without using BD cleanup is otherwise difficult/tedious. If you run into a situation where you have a significant portion of a diagram that is "static" and you want to manually lay it out in addition to a significant amount of scripted code you want to "auto layout", then I think the recommended practice is to extract all of the scripted code into a sub VI and use the VI boundary to isolate the cleanup actions.
> * There are probably more opportunities for further cleanup. For instance, there are cases where we update constants like string constants and don't re-layout because of that. This can lead to overlap/crowding on the block diagram. Similarly, there are cases where cluster constants on the block diagram only serve as a type specifier. In those cases, we could save space and the BD cleanup would be more effective by using the icon view for cluster. However, I think this setting gets lost when doing replacements, and I haven't taken the time to try and enforce it with scripting yet. Maybe in a future PR.
> 
> Anyway, I've audited generated code for both the client and the server, and I think these changes are an improvement in almost all cases. If you look at before and after comparisons of the generated code with this change, you can see for yourself. I couldn't think of a great way to easily show that with pictures, but I can attach some after pictures if there is something in particular you wanted to see.

I'm fine with it then since you've inspected the generated code.

Issue #339, opened 2024-01-13
Urgent: Memory is full


arevh:
When building large proto files LabVIEW stops the build process after several hours with the error:

LabVIEW: (Hex 0x2) Memory is full.
=========================
No Listeners on the GPIB.

I'm using LabVIEW 2020 64-bit, and at the time it brings the error message my RAM shows to be filles by 58%. My RAM is 20GB.
When this error pops up LabVIEW consumes approximately 1270MB of memory, so for sure for 64-bit LV and 20GB RAM, this is not really a memory issue.

gRPC version 1.0.1.1

AB#2624096

arevh:
Some more details which may be helpful. The Memory full error is shown for the following VI path:

Error 2 occurred at Invoke Node in gRPC Template Creation Utility-release.lvlib:Find Item in Library.vi->gRPC Template Creation Utility-release.lvlib:Get VIRef from Library.vi->gRPC Template Creation Utility-release.lvlib:Script Rich to Flat Conversion.vi->gRPC Template Creation Utility-release.lvlib:Create Oneof Message Converters (Rich to Flat).vi->gRPC Template Creation Utility-release.lvlib:Create Data Clusters for Types.vi->gRPC Template Creation Utility-release.lvlib:Main.vi

nokCHE:
We the similar issue here with large .proto files.

references are not closed in the "false" case inside vi 
"\gRPC\gRPC Server and Client Template [2]\gRPC Scripting Tools\Library API\Find Item in Library.vi"

closing the references solved this issue for us.
![image](https://github.com/user-attachments/assets/820ba8d8-3ca2-4b40-b39c-f79527035889)

please include fix in next version.

jasonmreding:
This should be resolved in https://github.com/ni/grpc-labview/releases/tag/v1.3.0.1

PR #401, opened 2025-01-08
Run grpc code generation in a private application context

jasonmreding:
# Description
Run grpc code generation in a private application context that is closed after each run of the generator.  This helps eliminate library/class leaks in the tools menu context which would otherwise require restarting LV to get rid of. It also makes it easier to test from source since the generator will no longer load template libraries into the source project while generating. Otherwise, LV tends to never unload libraries once loaded which often results in name collisions when running the generator multiple times on the same proto file. By loading everything into a private context, we avoid this since it acts like a project where everything gets unloaded from memory when the context is closed at the end.

# Implementation
For the most part this involved:
- Creating private application context at start of `Main.vi` and closing it before going idle
- Deleting application context terminals from existing VIs where it wasn't needed or didn't matter
- Adding application context terminals to connector panes of VIs where it is now needed and couldn't otherwise be extracted. In most cases, I was able to just retrieve the proper app context from an existing VI/library reference using `Owning App` property from the property node. I audited all usages of `Open VI Reference` and `New VI` nodes to ensure they all now use the appropriate app context and not just the "default" context.

In addition, I also:
- Fixed a few icons for correctness / readability as I noticed them.
- Refactored some duplicated code into a common sub VI. Otherwise, I would have had to make the same app context change to N cases of a case structure.
- Fixed and deleted work around for issue where sometimes we were getting error code 1 (mgArgErr) when trying to invoke the `Delete` method on a project item. This primarily happened when we try to cleanup template VIs that are no longer needed at the end of generation. This would typically manifest itself when generating both client and server at the same time or generating code for multiple proto files at the same time. I ultimately fixed this by invoking `ForceDelete` rather than `Delete`. I didn't fully debug the details of this. However, I believe what was happening is that the delete was removing the item from the library but wasn't actually removing/unloading the VI from memory. This then changed the qualified name of the VI since it is no longer part of the library. When doing the same thing on the second library, this would then create a name collision since both VIs have the same name when no longer qualified by a library. Calling `ForceDelete` works around this since it ignores the resulting name conflict. I also fixed some of the delete code that was only removing from the library and not deleting from disk.
- Cleaned up memory leak as described by #339 

# Testing
Manually tested generation of both the client and the server using proto APIs that included nested messages, enums, and oneof fields. Tested both Win32 and Win64 using LV 2019 and LV 2024.


labview source/Client Server Support New/gRPC Scripting Tools/Project API/Close Project.vi
dixonjoel, line 1: Can you explain why we *don't* want to pass the application reference here and use the default instead?
image

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Data Clusters for Types.vi
dixonjoel, line 1: I don't get why the icon is 'Create Type Meta' when the VI name is 'Create Data Clusters for Types'. I know you were just trying to make the icon more readable, but maybe we should change the text as well?
image

labview source/Client Server Support New/gRPC Scripting Tools/Class API/Get ControlRef from ClassItem.vi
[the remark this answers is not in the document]
jasonmreding: I don't know. I generally dislike and prefer not to check in disabled code unless there is an obvious reason and a comment. However, there are multiple spots throughout the generator code where there is disabled code that has been checked in without any explanation/comment to help explain why. Up to this point, I have left them alone rather than deleting them, and I haven't trying to trace back in history as to when they were added and possibly why.

labview source/Client Server Support New/gRPC Scripting Tools/Project API/Close Project.vi
jasonmreding: A project is effectively its own application context. More specifically, each target in the project forms its own application context. As such, it doesn't matter what app context (if any) is passed in. In the end, you still end up with a new app context for the project, and everything loaded in the project is in that context and not the app context passed in here.

labview source/Client Server Support New/gRPC Scripting Tools/VI API/Front Panel API/Create Type Definition.vi
dixonjoel, line 1: For the VIs where you added an app reference in, did you consider making it a pass-through and having app ref in / out? I don't know what the callers look like or if they use the app ref to call multiple sub-VIs, but it could clean up some calling code. Just a thought.

dixonjoel (approved):
Approved with a couple suggestions / questions.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Data Clusters for Types.vi
jasonmreding: If I had named this, I probably would have called it `Create Types from Metadata.vi` or something like that. However, the VI is actually doing more than that now. It's creating all of the typedefs plus the accessors for oneof classes and the rich <-> flat conversion VIs needed for oneof fields. This should probably be called `Generate Client RPC Messages.vi`. The existing `Generate RPC Messages.vi` should get renamed to `Generate Server RPC Messages.vi` if we go that route. I guess I can make that refactor as well if you like.

labview source/Client Server Support New/gRPC Scripting Tools/VI API/Front Panel API/Create Type Definition.vi
jasonmreding: I don't recall now any circumstances where it would have helped or I probably would have done it. If you saw any instances where chaining it through would help clean things up, please point those out.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Data Clusters for Types.vi
dixonjoel: Whatever you think is the best. I'm fine either way.

jasonmreding: I've refactored the VI names and icons as discussed above.

Issue #343, opened 2024-01-22
Neighboring Oneof structures issue


arevh:
If the message contains more than 1 Oneof structure, the second one's value is not converter properly from cluster value to grps message. Its value is returned as None.

In the example below dur_val will be received as None, even if some value was assigned in LabVIEW.

message WF_Constant {
  oneof value {
    string value_var_id = 2;
    float value_val = 3;
  }
  oneof duration {
    string dur_var_id = 5;
    float dur_val = 6;
  }
}

AB#2630427

jasonmreding:
This should be resolved in https://github.com/ni/grpc-labview/releases/tag/v1.3.0.1

PR #402, opened 2025-01-09
Fixing client serialization bugs

jasonmreding:
# Description
This addresses a number of client serialization issues around sparse enums and oneof fields. 
- In the case of sparse enums, we failed to translate back from the proto enum value to the LV enum value. This would result in the LV enum value always showing the last enum value. 
- For oneof fields we failed to assign the `selected_index` field in the cluster which made it look like no fields from the oneof were ever set when reading response data on the client. 
- This also fixes an issue where we sometimes were serializing the `selected_index` field across the wire which would end up getting captured in the unknown fields in the message on the other end.
- Messages containing more than one oneof field were not getting the `selected_index` field set correctly on their class cluster.

#396 
#343 

# Implementation
Most of these were only issues with the `LVMessageEfficient` derived class which I believe is only used for unary calls from the client. Otherwise, we end up using the `ClusterDataCopier` methods which already handled this.
- `LVMessage` now tracks data for which field of a oneof is selected. This data structure is updated when parsing a message. It is also updated by `ClusterDataCopier` when constructing a `LVMessage` from its cluster representation.
- `LVMessageEfficient` now assigns the `selected_index` field in the class cluster for a oneof during `PostInteralParseAction` using the new data structure added to `LVMessage`.
- `LVMessageEfficient` now has a hand written override for `ParseEnum` which translates from enum values from a proto file (which can be "sparse" or non-sequential) to the LV assigned enum value for the cluster since LV only supports sequential enum values.
- When assigning `selected_index` field values for the class cluster, we now iterate over `_elements` instead of `_mappedElements`. Since we use -1 as the field number for all `selected_index` fields, there will only ever be one entry in `_mappedElements`. This obviously doesn't work if the message has more than one oneof field.

# Testing
Manually tested generation of both the client and the server using proto APIs that included nested messages, enums, and oneof fields. Tested both Win32 and Win64 using LV 2019 and LV 2024.


src/cluster_copier.cc
```diff
@@ -16,28 +16,14 @@ namespace grpc_labview {
     // cluster: Pointer to the cluster created by LabVIEW
     void ClusterDataCopier::CopyToCluster(const LVMessage& message, int8_t* cluster)
     {
-        std::map<std::string, int> oneof_containerToSelectedIndexMap;
-        for (auto val : message._metadata->_mappedElements)
+        for (auto indexAndValue : message._values)
```
jasonmreding, line 19: Note: I reversed the order of the two loops here for efficiency. I also perform a map lookup rather than linear search which effectively eliminates the second loop.

src/cluster_copier.cc
```diff
@@ -16,28 +16,14 @@ namespace grpc_labview {
     // cluster: Pointer to the cluster created by LabVIEW
     void ClusterDataCopier::CopyToCluster(const LVMessage& message, int8_t* cluster)
     {
-        std::map<std::string, int> oneof_containerToSelectedIndexMap;
```
jasonmreding, line 19: Note: This container for tracking selected index of oneof containers has moved to the message class.

src/cluster_copier.cc
```diff
@@ -95,18 +81,22 @@ namespace grpc_labview {
...
-        // TODO: Skip the entire loop if the message has no oneof. It's a bool in the metadata.
-        for (auto val : message._metadata->_mappedElements)
+        // Second pass to fill the oneof selected_index. We can do this in one pass when we push the selected_field to the end of the oneof cluster!
+        if (message._oneofContainerToSelectedIndexMap.size() > 0)
         {
-            auto fieldMetadata = val.second;
-            if (fieldMetadata->isInOneof && fieldMetadata->protobufIndex < 0)
+            // Must iterate over _elements and not _mappedElements since all oneof selected_index fields use -1 for the field number
+            // and there can be multiple oneof fields in a message.
+            for (auto& fieldMetadata : message._metadata->_elements)
```
jasonmreding, line 89: Note: Need to traverse `_elements` and not `_mappedElements`.

src/cluster_copier.cc
```diff
@@ -116,31 +106,33 @@ namespace grpc_labview {
     //---------------------------------------------------------------------
     void ClusterDataCopier::CopyFromCluster(LVMessage& message, int8_t* cluster)
     {
-        message._values.clear();
-        std::map<std::string, int> oneof_containerToSelectedIndexMap; // Needed to serialize only the field related to the selected_index
-        for (auto val : message._metadata->_mappedElements)
+        message.Clear();
+        for (auto& fieldMetadata : message._metadata->_elements)
```
jasonmreding, line 110: Note: Need to traverse `_elements` and not `_mappedElements`.

src/lv_message_efficient.h
```diff
@@ -47,34 +48,36 @@ namespace grpc_labview
...
-        const char* ParseFixed32(const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
-        const char* ParseFixed64(const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
-        const char* ParseSFixed32(const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
-        const char* ParseSFixed64(const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
-        const char* ParseString(unsigned int tag, const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
-        const char* ParseBytes(unsigned int tag, const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
-        const char* ParseNestedMessage(google::protobuf::uint32 tag, const MessageElementMetadata& fieldInfo, uint32_t index, const char* ptr, google::protobuf::internal::ParseContext* ctx);
+        int8_t* _LVClusterHandle;
```
jasonmreding, line 51: Note: I moved book keeping for the cluster pointer from `LVMessage` base class to here as it was only ever used by this class. The cluster pointer is now required to be passed during construction rather than exposing a public set method you have to call after the fact. I also changed the type from `const char*` to `int8_t*` as that generally results in the least amount of type conversion logic around the pointer through the various APIs in which it is used.

src/lv_message_efficient.cc
```diff
@@ -57,7 +57,6 @@ namespace grpc_labview
     DEFINE_PARSE_FUNCTION(uint64_t, UInt64, UINT64, UInt64)
     DEFINE_PARSE_FUNCTION(float, Float, FLOAT, Float)
     DEFINE_PARSE_FUNCTION(double, Double, DOUBLE, Double)
-    DEFINE_PARSE_FUNCTION_SPECIAL(int32_t, Enum, ENUM, Enum)
```
jasonmreding, line 60: Note: There is now a custom implementation of this rather than using this macro so we can perform translation between proto enum value and LV enum values.

src/lv_message.cc
```diff
@@ -105,8 +106,16 @@ namespace grpc_labview
...
                 {
-                    auto fieldInfo = (*fieldIt).second;
+                    auto& fieldInfo = (*fieldIt).second;
                     LVMessageMetadataType dataType = fieldInfo->type;
+
+                    if (fieldInfo->isInOneof)
+                    {
+                        // set the map of the selected index for the "oneofContainer" to this protobuf Index
+                        assert(_oneofContainerToSelectedIndexMap.find(fieldInfo->oneofContainerName) == _oneofContainerToSelectedIndexMap.end());
+                        _oneofContainerToSelectedIndexMap.insert(std::pair<std::string, int>(fieldInfo->oneofContainerName, fieldInfo->protobufIndex));
```
bkeryan, line 116: nit: You should be able to use an initializer list here instead of spelling out the type.

```suggestion
                        _oneofContainerToSelectedIndexMap.insert({ fieldInfo->oneofContainerName, fieldInfo->protobufIndex });
```

src/cluster_copier.cc
bkeryan, line 19: nit: I think you can save another addref/release here.
```suggestion
        for (auto& indexAndValue : message._values)
```

src/cluster_copier.cc
```diff
@@ -116,31 +106,33 @@ namespace grpc_labview {
...
-                {
-                    // set the map of the selected index for the "oneofContainer" to this protobuf Index
-                    assert(oneof_containerToSelectedIndexMap.find(fieldMetadata->oneofContainerName) == oneof_containerToSelectedIndexMap.end());
-                    auto selected_index = *(int*)(cluster + fieldMetadata->clusterOffset);
-                    oneof_containerToSelectedIndexMap.insert(std::pair<std::string, int>(fieldMetadata->oneofContainerName, selected_index));
-                }
+                // set the map of the selected index for the "oneofContainer" to this protobuf Index
+                assert(message._oneofContainerToSelectedIndexMap.find(fieldMetadata->oneofContainerName) == message._oneofContainerToSelectedIndexMap.end());
+                auto selected_index = *(int*)(cluster + fieldMetadata->clusterOffset);
+                message._oneofContainerToSelectedIndexMap.insert(std::pair<std::string, int>(fieldMetadata->oneofContainerName, selected_index));
```
bkeryan, line 117: nit: You should be able to use an initializer list here instead of spelling out the type.

```suggestion
                message._oneofContainerToSelectedIndexMap.insert({ fieldMetadata->oneofContainerName, selected_index });
```

src/lv_message_efficient.h
```diff
@@ -18,11 +18,12 @@ namespace grpc_labview
     class LVMessageEfficient : public LVMessage
     {
     public:
-        LVMessageEfficient(std::shared_ptr<MessageMetadata> metadata) : LVMessage(metadata) {}
+        LVMessageEfficient(std::shared_ptr<MessageMetadata> metadata, int8_t* cluster) : LVMessage(metadata) { _LVClusterHandle = cluster; }
```
bkeryan, line 21: nit: Use initializers

```suggestion
        LVMessageEfficient(std::shared_ptr<MessageMetadata> metadata, int8_t* cluster) : LVMessage(metadata), _LVClusterHandle(cluster) {}
```

src/lv_message_efficient.h
```diff
@@ -93,7 +96,7 @@ namespace grpc_labview
             if (numElements != 0)
             {
                 auto messageTypeSize = sizeof(MessageType);
-                NumericArrayResize(GetTypeCodeForSize(messageTypeSize), 1, reinterpret_cast<void*>(const_cast<char*>(_lv_ptr)), numElements);
+                NumericArrayResize(GetTypeCodeForSize(messageTypeSize), 1, reinterpret_cast<void*>(_lv_ptr), numElements);
```
bkeryan, line 99: FYI: I don't think you need to explicitly cast to `void*`, since the NumericArrayResize wrapper accepts `void*` and `int8_t*` is convertible to `void*`.

src/lv_message_efficient.cc
```diff
@@ -67,10 +66,46 @@ namespace grpc_labview
...
+
+        if (fieldInfo.isRepeated)
+        {
+            auto repeatedEnum = std::make_shared<LVRepeatedEnumMessageValue>(index);
+            ptr = PackedEnumParser(&(repeatedEnum->_value), ptr, ctx);
+
+            int count = repeatedEnum->_value.size();
+            for (size_t i = 0; i < count; i++)
+            {
+                auto enumValueFromPrtobuf = repeatedEnum->_value[i];
```
bkeryan, line 82: spelling: Prtobuf, here and below

src/lv_message_efficient.cc
```diff
@@ -173,24 +208,44 @@ namespace grpc_labview
...
+        {
+            // Must iterate over _elements and not _mappedElements since all oneof selected_index fields use -1 for the field number
+            // and there can be multiple oneof fields in a message.
+            for (auto& fieldMetadata : _metadata->_elements)
+            {
+                if (fieldMetadata->isInOneof && fieldMetadata->protobufIndex < 0)
+                {
+                    // This field is the selected_index field of a oneof. This field only exists in the cluster
+                    // and is not a field in the message.
+                    if (_oneofContainerToSelectedIndexMap.find(fieldMetadata->oneofContainerName) != _oneofContainerToSelectedIndexMap.end())
```
bkeryan, line 232: This does an extra map lookup with operator[] below. Save the iterator and use it in the assignment instead.

FYI, there's similar code in CopyToCluster, maybe it could be extracted into an LVMessage method.

bkeryan (changes requested):

bkeryan (approved):

PR #404, opened 2025-01-28
Adding support for double 2D arrays

jasonmreding:
# Description
This PR adds built in support for 2D arrays of doubles based on [array.proto](https://github.com/ni/ni-apis/blob/main/ni/protobuf/types/array.proto). With this change, authors of grpc services and clients of the service can utilize the built in type for 2D arrays of doubles on the diagram without having to convert to/from the equivalent grpc message type that gets sent across the wire. Instead, the conversion is handled internally for the user. This PR is meant to serve as a template for subsequent "well known" types that allow users to utilize native LV types that aren't inherently supported by grpc without requiring them to generate code from a proto file.

# Implementation / Design
There were a number of implementation options that were considered, that I ultimately rejected in favor of other approaches. Specifically, it might seem odd that there is no generated code from either the C++ plugin to protoc or from grpc-labview itself as part of this change. However, I ultimately rejected that approach as the C++ code currently utilizes a generic `LVMessage` class/API for all of its serialization. While the generated message classes inherit from the same base class, I didn't see a good way to make that work with the current design. 

Similarly, I also considered just generating conversion code at the LV layer via our generator similar to how oneof fields are handled. While this would work for the standard client/service generation, it wouldn't work for the <vilib> APIs that integrate with the `Any` type since there is no calls into generated code with those APIs. That means we would either have to duplicate the logic for these APIs or just not support "well known" types with these APIs.

Here are the high level changes to each part of the code:
C++
- Introduced a well known types enum and namespace. Usages of the enum can be searched to quickly find most places that have to be modified when adding a new well known type.
- All of the data type traversal code paths that visit nested clusters / messages have been updated to first check if the field is a well known type. If so, a custom method is called to handle serialization for each supported well known type. If the message isn't recognized as a well known type, it falls back to the existing handling of nested messages.
- There is now a global / singleton `IMessageElementMetadataOwner` for well known types. This instance is responsible for registering metadata for well known types and avoids users having to specifically generate code for these types and registering the metadata themselves. In practice, this is really only utilized by the `Register Cluster.vi` from <vilib>. For generated clients/servers, we still generate metadata for the type at the LV that gets registered with the generated client/server code. When this happens, the metadata registered with the client/server takes precedence over the global version.

LV
- Introduced `Well Known Types.ctl` typedef to `grpc-lvsupport.lvlib`. Usages of this typedef can be searched to quickly find most places that have to be modified when adding a new well known type.
- There were a number of spots in the generator that had to be updated to check if a field is a well known type. If it is, the generator will not treat the field as a nested message that needs to be traversed into, and it won't try to generate a typedef for the field. There were often separate code paths for "regular" fields and fields that were part of a oneof field that had to be updated to account for this.
- There was a lot of refactoring / cleanup to VIs that adds a lot of noise to the core changes. I will attempt to call out attention to the most important changes with comments on individual files.

With the above in mind, the following use cases are not supported with this PR:
- A repeated field of `Double2DArray` is not recognized as a well known type. Instead, it will use the generated message cluster the same as it does today. Since LV doesn't support arrays of arrays, we would need to implicitly convert the diagram types to array of cluster of 2D array similar to how repeated byte fields are handled. This is doable, but I didn't spend time on it as part of this PR.
- `Double2DArray` is only recognized as a field in a message and not as the top level request/response message itself. So if you define a RPC in a proto file where the request/response message is `Double2DArray`, you will still have to deal with the cluster representation of the type on the diagram and not the built in 2D array type.
- The LV <vilib> APIs for packing/unpacking messages a field at a time do not support well known types. They currently don't support any message types, and I am not improving that as part of this PR. The pack/unpack APIs for an entire message work, but those require metadata registration where as the packing/unpacking a field at a time do not.

# Testing
I manually tested code generation of client and server and round tripping of data with the following proto file.

```protobuf
syntax = "proto3";

package echo.array;

import "ni/protobuf/types/array.proto";

service EchoArray {
  rpc Echo2DArray(Echo2DArrayRequest) returns(Echo2DArrayResponse);
  rpc EchoRepeated2DArray(EchoRepeated2DArrayRequest) returns(EchoRepeated2DArrayResponse);
}

message NestedMessage {
  oneof union
  {
    string string_value = 1 ;
    ni.protobuf.types.Double2DArray double_2d_array = 2;
    NestedMessage2 nested_message_2 = 3;
  }
}

message NestedMessage2 {
  int32 int_value = 1;
}

message Echo2DArrayRequest {
  bool bool_value = 1 ;
  ni.protobuf.types.Double2DArray array1 = 2;
  NestedMessage nested_message = 3;
  ni.protobuf.types.Double2DArray array2 = 4;
}

message Echo2DArrayResponse {
  bool bool_value = 1;
  ni.protobuf.types.Double2DArray array1 = 2;
  NestedMessage nested_message = 3;
  ni.protobuf.types.Double2DArray array2 = 4;
}

message EchoRepeated2DArrayRequest {
  repeated ni.protobuf.types.Double2DArray repeated_array = 1;
}

message EchoRepeated2DArrayResponse {
  repeated ni.protobuf.types.Double2DArray repeated_array = 1;
}
```

I also verified the following cluster packs/unpacks correctly.

image

# Reviewers
- Adding @bkeryan to focus on feedback of C++ changes.
- Adding @dixonjoel to focus on feedback of LV changes.
- Adding @ccaltagi  as an FYI since I know he was working on similar changes for WDT.
- Adding @ccifra  for any feedback on overall design / approach taken.


labview source/Client Server Support New/Oneof Template/GetActiveFieldName.vi
jasonmreding, line 1: GetFieldFromProtobufIndex.vi was renamed to this for consistency and an easier to understand API that the user consumes.

labview source/gRPC lv Support/Server API/Cluster to gRPC message/Register Cluster.vi
jasonmreding, line 1: Note: All of the changes under this library are to support this public <vilib> VI. This library is a runtime library which is a dependency of our generator.

labview source/gRPC lv Support/Shared/Well Known Types.ctl
jasonmreding, line 1: Note: This is primary typedef used to define well known types. You can follow usages of this typedef to find all the spots in the code that need to account for well known types in some fashion.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Flat to Rich Oneof Conversion.vi
jasonmreding, line 1: Note: VI was renamed for consistency. Common code was then extracted into `Script Field Conversion.vi`.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Rich to Flat Oneof Conversion.vi
jasonmreding, line 1: Note: VI was renamed for consistency. Common code was then extracted into `Script Field Conversion.vi`.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Create Enum Data Clusters (simple).vi
jasonmreding, line 1: Note: There was code here that would check is an already existing typedef (generate from proto file again) still matches the .proto file definition. If it didn't, it would delete all of the controls and start over. Rather than updating the matching code to account for well known types, I just always delete any existing controls on the typedef and always start from scratch. That seems much simpler, and I don't think there will much difference in terms of performance.

labview source/gRPC lv Support/Server API/Cluster to gRPC message/Cluster to Element Details.vi
jasonmreding, line 1: Note: Refactored to push more shared logic into sub VIs. Part of the refactoring accounts for well known types.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Update Control for Element.vi
jasonmreding, line 1: Note: functional change to support well known types

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Is Element Conversion Required.vi
jasonmreding, line 1: Note: Functional change to support well known types. This indicates whether oneof conversion VIs are required to be scripted when accessing the field.

labview source/Client Server Support New/gRPC Scripting Tools/Top Level API/Is Element Well Known Type.vi
jasonmreding, line 1: Note: functional change to support well known types

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Update OneOf Field Accessor.vi
jasonmreding, line 1: Note: functional change to support well known types

labview source/Client Server Support New/gRPC Scripting Tools/VI API/Front Panel API/Add TypeDef to Cluster.vi
jasonmreding, line 1: Note: functional change to support well known types

labview source/gRPC lv Support/Server API/Cluster to gRPC message/Element to Message Element Metadata.vi
jasonmreding, line 1: Note: functional change to support well known types

src/grpc_interop.cc
```diff
@@ -36,66 +36,6 @@ namespace grpc_labview
         DSDisposeHandle(eventData.methodName);
     }
 
-    //---------------------------------------------------------------------
```
jasonmreding, line 39: Note: This code as moved into the constructor of the `LVMessageMetadata` class itself as that seemed more appropriate division of responsibility.

src/lv_message_efficient.cc
```diff
@@ -215,6 +221,19 @@ namespace grpc_labview
         return protobuf_ptr;
     }
 
+    //---------------------------------------------------------------------
+    //---------------------------------------------------------------------
+    const char* LVMessageEfficient::ParseDouble2DArrayMessage(google::protobuf::uint32 tag, const MessageElementMetadata& fieldInfo, uint32_t index, const char* protobuf_ptr, ParseContext* ctx)
+    {
+        auto metadata = fieldInfo._owner->FindMetadata(fieldInfo.embeddedMessageName);
+        auto nestedMessage = std::make_shared<LVMessage>(metadata);
```
jasonmreding, line 229: Note: Here we create a regular `LVMessage` to copy the type. This is a bit clunky from an efficiency perspective, but I could think of a way to do it without copying into an intermediate message value since there is no guarantee on the order in which fields are sent over the wire.

labview source/Client Server Support New/Oneof Template/GetActiveFieldName.vi
dixonjoel, line 1: I assume these arrays are populated as part of the code generation process?
image

dixonjoel:
I'm seeing some broken VIs, but I'm not sure if I am using the right version of LV for the diffs or if I have the right things installed. Is 'To Well Known Type.vi' supposed to be part of this PR?
image

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Field Conversion.vi
dixonjoel, line 1: I'm seeing some broken VIs, but I'm not sure if I am using the right version of LV for the diffs or if I have the right things installed. Is 'To Well Known Type.vi' supposed to be part of this PR?
image

Or is there an incorrect linkage to vi.lib?

jasonmreding: Yeah, the VI is being added as part of this and is part of  `grpc-lvsupport.lvlib` which is part of the `LabVIEW gRPC Library` VIPM package. The `gRPC Server and Client Template [2]` VIPM package contains all of the generator code, and this package depends on the  `LabVIEW gRPC Library` package. That means you generally have to have the  `LabVIEW gRPC Library` package installed to view/edit the generator code. That's typically not a problem because you probably already have a prior version of that package installed. However, it is a bit of a PITA when adding a new API and then bubbling up changes on top of those new APIs. If that is a nuisance for the review, there really isn't really a good way to deal with it that I'm aware of other than to build the package locally and then install it from your local build. Then don't forget to revert vipb version bump changes and uninstall afterwards (or downgrade) if you care about keeping your system to with only publicly published packages.

labview source/Client Server Support New/Oneof Template/GetActiveFieldName.vi
jasonmreding: Yes, there is scripting code that looks for constants with those labels and updates them based on the metadata. In fact, prior to this PR there was a bit of a bug where it was getting populated twice. You can look at existing clients/servers to see what it looks like. If you search for the label text used for those constants, you should see the VI where it is getting updated.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Field Conversion.vi
jasonmreding: I guess that means I should probably update the version number in the .vipb for both packages and updated the oldest allowed dependency for the generator package. I'm thinking bumping to 1.3.0.1 probably makes more sense than 1.2.8.1. Thoughts?

dixonjoel: Seems like a pretty significant change, so I'd say a minor version bump is in order. So 1.3.0.1 is my vote.

dixonjoel: And can you explain where the 'LabVIEW gRPC Servicer' package fits into what you explained above?
image

bkeryan (changes requested):
I'm not through reviewing this yet.

dixonjoel:
There's a bunch of projects still saving in LV 2019. Have you talked to anyone about possibly updating those to a newer version of LV?

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Update OneOf Field Accessor.vi
dixonjoel, line 1: Check for understanding. Does the fact that there's no 'Double2DArray' case here indicate that it's not supported in a 'oneof'? I'm guessing we've never supported messages as a oneof element?
image

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Field Conversion.vi
jasonmreding: I haven't really looked into that part of the code much, but from what I can see it manages runtime libraries deployed to <vilib>. It effectively defines the LV classes/interfaces used by the generated server side code that implements the service API. So it is effectively equivalent the `NI Measurement Plug-In SDK.lvlib` library used for measurement plugins. So:
- LabVIEW grpc Library
  - Deployed to <vilib> and has its own top level functions palette 
  - Provides generic APIs for interacting with gRPC (open client connection, pack/unpack, etc.)
  - Also deploys gRPC dlls built from C++ code in this repo
- LabVIEW gRPC Servicer 
  - Depends on LabVIEW gRPC Library
  - Deployed to <vilib> and has its own top level functions palette (I'm not sure why it is on the palette)
  - Defines base classes/interfaces for generated gRPC server implementation.
  - All LV code
- gRPC Server and Client Template [2] (I don't know what the [2] in the name is about)
  - Depends on LabVIEW gRPC Library and LabVIEW gRPC Servicer
  - Deployed to <project> folder so it shows up in tools menu
  - All LV code

At least that is my understanding of things.

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Update OneOf Field Accessor.vi
jasonmreding: So for a oneof field, we end up generating a LV class from template where every field in the oneof has a "Get", "Set", and "Has" <field name> template VI. Up to this point, the scripting code has already done a first pass of the scripting which converts control/indicator for the field value to the appropriate control/data type. This is second pass that assigns and fixes up the controls to the appropriate typedef reference (since the typedef might not have been generated yet when the first pass occurred). Since message and enum fields are the only fields that will correspond to a typedef, they are the only thing handled here. While a "well known" type is a message type from a gRPC perspective, we are mapping it to a built in LV type and not a typedef. That is why we skip all well known types here since they should map to an intrinsic type and not a custom typedef type. At least that is the idea at the moment. Perhaps in the future we will want to introduce a "well known" type where this doesn't hold, and we will have to figure out how to handle and classify that type at that point in time.

jasonmreding:
> There's a bunch of projects still saving in LV 2019. Have you talked to anyone about possibly updating those to a newer version of LV?

I haven't talked to anybody, but I think there are dragons here. From my experience, there are lots of test suite and other build image assumptions about use of 32 bit LV 2019 that changing the LV version will be non-trivial. This should really be cleaned up at some point, but I believe this will be a non-trivial amount of tech debt to address for what otherwise seems like it should be a simple repo change.

dixonjoel (approved):

labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Script Field Conversion.vi
dixonjoel: I built the packages and everything's unbroken now.

src/message_metadata.cc
```diff
@@ -0,0 +1,124 @@
...
+        auto name = GetLVString(lvMetadata->messageName);
+        messageName = name;
+        typeUrl = GetLVString(lvMetadata->typeUrl);
+        clusterSize = 0;
+        alignmentRequirement = 0;
+
+        if (lvMetadata->elements != nullptr)
+        {
+            // byteAlignment for LVMesageElementMetadata would be the size of its largest element which is a LStrHandle
+            auto lvElement = (LVMesageElementMetadata*)(*lvMetadata->elements)->bytes(0, sizeof(LStrHandle));
```
bkeryan, line 107: Should this use `alignof(LVMesageElementMetadata)`? https://en.cppreference.com/w/cpp/language/alignof

src/message_metadata.cc
```diff
@@ -0,0 +1,124 @@
...
+#include "message_metadata.h"
+#include "lv_interop.h"
+#include "well_known_types.h"
+
+namespace grpc_labview
+{
+    //---------------------------------------------------------------------
+    // Functions for calculating whether an element is considered a well known type.
+    //---------------------------------------------------------------------
+    static wellknown::Types IsWellKnownDouble2DArray(const MessageElementMetadata& metadata)
```
bkeryan, line 12: IsWellKnownDouble2DArray sounds like a predicate that returns bool. I recommend renaming this to something like GetDouble2DArrayWellKnownTypeEnum or converting it to a lambda so you can include it in the map initializer.

e.g. 
```
[](const MessageElementMetadata& metadata) {
        if (!metadata.isRepeated)
        {
            return wellknown::Types::Double2DArray;
        }
        return wellknown::Types::None;
}
```

bkeryan (changes requested):

src/message_metadata.h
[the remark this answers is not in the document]
jasonmreding: Fixed spelling and updated test that checks for backward compatibility of struct used to represent LV cluster.

src/lv_interop.h
```diff
@@ -96,6 +96,34 @@ namespace grpc_labview
     using LV1DArrayPtr = LV1DArray*;
     using LV1DArrayHandle = LV1DArray**;
 
+    struct LV2DArray {
+        int32_t firstDimensionSize; /* number of T elements for the first dimension */
+        int32_t secondDimensionSize; /* number of T elements for the second dimension */
```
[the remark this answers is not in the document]
jasonmreding: I went with `int32_t dimensionSizes[2]`.

src/lv_interop.h
```diff
@@ -96,6 +96,22 @@ namespace grpc_labview
     using LV1DArrayPtr = LV1DArray*;
     using LV1DArrayHandle = LV1DArray**;
 
+    struct LV2DArray {
+        int32_t dimensionSizes[2]; /* number of T elements for each dimension */
+        int8_t rawBytes[1]; /* (firstDimensionSize * secondDimensionSize * sizeof(T)) bytes */
+
+        // The start of 2D array data is always aligned and does not require padding.
```
bkeryan, line 103: This is true for 1/2/4/8 but not 16+. Is it possible to static_assert that alignof(T) <= alignof(LV2DArray)?

src/lv_interop.h
```diff
@@ -96,6 +96,34 @@ namespace grpc_labview
...
+        template<typename T>
+        T* bytes(int byteOffset)
+        {
+            static_assert(!std::is_class<T>::value, "T must not be a struct/class type.");
+            return (T*)(bytes(byteOffset, sizeof(T)));
+        }
+
+        void* bytes(int byteOffset, int byteAlignment)
+        {
+            return (void*)(rawBytes + AlignClusterOffset(4, byteAlignment) - 4 + byteOffset);
```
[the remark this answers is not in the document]
jasonmreding: OK, I think I understand the original intent of the code better now. For 1D arrays on a 64 bit OS, there will be a 4 byte pad between the dimension size field and the start of the array data pointer to keep the data aligned at 8 bytes. For 32 bit OS, no padding is used since 4 byte alignment works for all primitive array data types (at least the ones grpc cares about).

For 2D arrays, no padding is ever needed for alignment purposes between the 32 bit dimension sizes and the start of the array data. This is true for both 32 bit and 64 bit OS. This is based on observations of the raw memory layout applied by LV in the debugger as well as [this](https://www.ni.com/docs/en-US/bundle/labview/page/how-labview-stores-data-in-memory.html#d25151e190) and [this](https://www.ni.com/docs/en-US/bundle/labview/page/using-arrays-and-strings-in-the-call-library-function-node.html#d69997e265).

Based on this, I'm now only providing a `bytes()` API since no alignment calculation is needed, and I never need to calculate offsets into the array (which would only really be needed for 2D array of clusters).

src/lv_interop.h
bkeryan: We discussed this offline, I think asserting `sizeof(T) <= 8` is fine

src/well_known_types.h
```diff
@@ -0,0 +1,69 @@
...
+        // https://github.com/ni/ni-apis/blob/main/ni/protobuf/types/array.proto
+        //---------------------------------------------------------------------
+        class Double2DArray
+        {
+        public:
+            static const std::string& GetMessageName();
+            static const std::string& GetMessageUrl();
+
+            static void CopyToCluster(const grpc_labview::MessageElementMetadata& metadata, int8_t* start, const std::shared_ptr<LVMessageValue>& value);
+            static void CopyFromCluster(const std::shared_ptr<MessageElementMetadata> metadata, int8_t* start, LVMessage& message);
```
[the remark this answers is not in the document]
jasonmreding: Honestly, I didn't give this much thought. I just copied the same signatures for similar functions defined in `cluster_copier.h`. I think the intent was to disallow modification to the element metadata though.

src/well_known_types.cc
```diff
@@ -0,0 +1,177 @@
...
+            {
+                return;
+            }
+
+            int rows = (*array)->firstDimensionSize;
+            int columns = (*array)->secondDimensionSize;
+            if (rows == 0 || columns == 0)
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
bkeryan: We discussed this offline. This is true for any message field, and the behavior is language-dependent. In Python, it's like LV and nested messages always exist, even when unset. In C# and C++, clients may have to check whether the nested message exists. 

I'm closing this as resolved.

src/well_known_types.h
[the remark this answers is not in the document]
jasonmreding: OK, I've updated all of the cluster copy related methods to have `std::shared_ptr<const MessageElementMetadata> metadata` and `std::shared_ptr<const LVMessageValue>& value` parameters.

src/well_known_types.cc
```diff
@@ -0,0 +1,177 @@
...
+                owner->FinalizeMetadata();
+            }
+            return *_instance;
+        }
+
+        //---------------------------------------------------------------------
+        //---------------------------------------------------------------------
+        MetadataOwner::MetadataOwner()
+        {
+            _owner = new MessageElementMetadataOwner();
```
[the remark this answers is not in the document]
jasonmreding: Updated to use `std::unique_ptr`.

src/message_metadata.cc
jasonmreding: Updated.

src/message_metadata.cc
jasonmreding: Updated to `CalculateDouble2DArrayWellKnownTypeEnum`.

src/lv_interop.h
jasonmreding: Added assert for size of element being <= 8

src/well_known_types.cc
```diff
@@ -0,0 +1,177 @@
...
+
+            auto elementCount = rows * columns;
+            if (elementCount > 0)
+            {
+                NumericArrayResize(GetTypeCodeForSize(sizeof(double)), 2, start, elementCount);
+                auto array = *(LV2DArrayHandle*)start;
+                (*array)->firstDimensionSize = rows;
+                (*array)->secondDimensionSize = columns;
+                auto byteCount = elementCount * sizeof(double);
+                memcpy((*array)->bytes<double>(), dataValue->_value.data(), byteCount);
```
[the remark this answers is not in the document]
jasonmreding: Talked off line about this, but to properly error report this would require a lot of rework with the existing code base. That level of change probably isn't appropriate to add on top of what is already a pretty big change. Instead, I've added checks to prevent buffer overwrite without throwing any errors. If the amount of array data written with the message is less than indicated by the array dimensions, we still allocate the larger array size in LV but write less data. If there is more data in the array, then we simply truncate the amount of data copied to what is indicated by the array dimensions from the message. I have tested both cases in the debugger to verify it works as expected. If we write less data than expected, then the unused/unwritten portion of the array shows up as zero in LV.

src/cluster_copier.cc
```diff
@@ -17,9 +17,9 @@ namespace grpc_labview {
     // cluster: Pointer to the cluster created by LabVIEW
     void ClusterDataCopier::CopyToCluster(const LVMessage& message, int8_t* cluster)
     {
-        for (auto& indexAndValue : message._values)
+        for (const auto& indexAndValue : message._values)
         {
-            auto& value = indexAndValue.second;
+            std::shared_ptr<const LVMessageValue> value = indexAndValue.second;
```
bkeryan, line 22: Oops, this is an unintended consequence of changing `const shared_ptr<T>&` to `shared_ptr<U>&`.

`const shared_ptr<T>&` accepts a const ref to a shared pointer, so it allows conversion from another shared_ptr type due to subclassing/constness of T. Doing a conversion constructs a temporary shared_ptr and a const ref is allowed to bind to a temporary.

`shared_ptr<U>&` only accepts a mutable ref to `shared_ptr<U>` exactly, with no conversions. Mutable refs are not allowed to bind to temporaries.

If you change these functions to accept `const std::shared_ptr<const LVMessageValue>&` (const ref to shared_ptr to const T) or `std::shared_ptr<const LVMessageValue>` (pass by value shared_ptr to const T) then you won't have to declare a separate local variable in order to have a mutable shared_ptr to pass to the function.

src/lv_message_efficient.cc
```diff
@@ -230,7 +230,8 @@ namespace grpc_labview
         protobuf_ptr = ctx->ParseMessage(nestedMessage.get(), protobuf_ptr);
         auto nestedClusterPtr = _LVClusterHandle + fieldInfo.clusterOffset;
         auto nestedMessageValue = std::make_shared<LVNestedMessageMessageValue>(index, nestedMessage);
-        wellknown::Double2DArray::CopyToCluster(fieldInfo, nestedClusterPtr, nestedMessageValue);
+        auto constNestedMessageValue = std::static_pointer_cast<const LVMessageValue>(nestedMessageValue);
```
bkeryan, line 233: This is the same thing as in cluster_copier.cc: `shared_ptr<LVNestedMessageMessageValue>` should be implicitly convertible to `shared_ptr<const LVMessageValue>`, but because the function expects a mutable ref you have to declare a separate local variable on the stack.

If `wellknown::Double2DArray::CopyToCluster` takes a `shared_ptr<const LVMessageValue>` by value or `const shared_ptr<const LVMessageValue>&` by const reference then conversions will be allowed and you won't need to declare a local variable like this.

src/cluster_copier.cc
jasonmreding: Updated to `const std::shared_ptr<const LVMessageValue>&` and `const std::shared_ptr<const MessageElementMetadata>` throughout.

src/lv_message_efficient.cc
jasonmreding: Updated to `const shared_ptr<const LVMessageValue>&`.

bkeryan (approved):
The C++ code looks good to me. I did not review the G code.

PR #405, opened 2025-01-31
Update package versions

jasonmreding:
# Description
Increment package versions to indicate new features.

# Implementation
Updated to new minor version to indicate presence of new features. Updated minimum package version since we are using new APIs that are only present in the latest package.

# Testing
N/A


dixonjoel:
Do we not want to bump https://github.com/ni/grpc-labview/blob/master/labview%20source/gRPC%20lv%20Servicer/build%20spec/LabVIEW%20gRPC%20Servicer.vipb as well?

dixonjoel (approved):
Approved with a query about another .vipb.

jasonmreding:
> Do we not want to bump https://github.com/ni/grpc-labview/blob/master/labview%20source/gRPC%20lv%20Servicer/build%20spec/LabVIEW%20gRPC%20Servicer.vipb as well?

There is nothing in that package that changed which is why I didn't change it. This would be my preference over say always bumping all three packages to have the same version number even if nothing changed in some of the packages.

dixonjoel:
> > Do we not want to bump https://github.com/ni/grpc-labview/blob/master/labview%20source/gRPC%20lv%20Servicer/build%20spec/LabVIEW%20gRPC%20Servicer.vipb as well?
> 
> There is nothing in that package that changed which is why I didn't change it. This would be my preference over say always bumping all three packages to have the same version number even if nothing changed in some of the packages.

Maybe. But then when it comes to making a release, you'd release packages with different versions? That seems confusing to me. Anyway, we can defer this decision to a release.

PR #407, opened 2025-02-07
Adding support for 2D string arrays similar to 2D double arrays.

jasonmreding:
# Description
This PR is a follow on to #404 and adds built in support for 2D arrays of strings based on [array.proto](https://github.com/ni/ni-apis/blob/main/ni/protobuf/types/array.proto). With this change, authors of grpc services and clients of the service can utilize the built in type for 2D arrays of strings on the diagram without having to convert to/from the equivalent grpc message type that gets sent across the wire. Instead, the conversion is handled internally for the user. 

# Implementation / Design
C++:
- Split `wellknown::Types` enum and well known message classes into separate files. This avoided some circular dependencies in the header files and eliminates the need to forward declare types instead of including the header directly to break the cycle.
- Created base class and interface class for 2D message classes to better single source common code. Most of the static methods are now virtual functions, and there is a singleton `GetInstance` method which is used instead of static methods.
- Introduced `String2DArray` message class.
- Most everything else is just following usages of `wellknown::Types` enum and updating the code as appropriate.

LV:
- Added new String2DArray enum value to `Well Known Types.ctl`
- Most everything else was following usages of the enum type and updating accordingly and/or fixing broken case structures that no longer had a case for every enum value. To make it easier to find spots in the code that need to be revisited when adding a new well known type, I removed default case structure cases that returned a runtime error. This means you can utilize compile time errors rather than runtime errors when fixing up the code.

# Testing
I manually tested code generation of client and server and round tripping of data with the following proto file.

```protobuf
syntax = "proto3";

package echo.array;

import "ni/protobuf/types/array.proto";

service EchoArray {
  rpc Echo2DArray(Echo2DArrayRequest) returns(Echo2DArrayResponse);
  rpc EchoRepeated2DArray(EchoRepeated2DArrayRequest) returns(EchoRepeated2DArrayResponse);
}

message NestedMessage {
  oneof union
  {
    string string_value = 1 ;
    ni.protobuf.types.Double2DArray double_2d_array = 2;
    NestedMessage2 nested_message_2 = 3;
    ni.protobuf.types.String2DArray string_2d_array = 4;
  }
}

message NestedMessage2 {
  int32 int_value = 1;
}

message Echo2DArrayRequest {
  bool bool_value = 1 ;
  ni.protobuf.types.Double2DArray array1 = 2;
  NestedMessage nested_message = 3;
  ni.protobuf.types.Double2DArray array2 = 4;
  ni.protobuf.types.String2DArray string_array = 5;
}

message Echo2DArrayResponse {
  bool bool_value = 1;
  ni.protobuf.types.Double2DArray array1 = 2;
  NestedMessage nested_message = 3;
  ni.protobuf.types.Double2DArray array2 = 4;
  ni.protobuf.types.String2DArray string_array = 5;
}

message EchoRepeated2DArrayRequest {
  repeated ni.protobuf.types.Double2DArray repeated_double_array = 1;
  repeated ni.protobuf.types.String2DArray repeated_string_array = 2;
}

message EchoRepeated2DArrayResponse {
  repeated ni.protobuf.types.Double2DArray repeated_double_array = 1;
  repeated ni.protobuf.types.String2DArray repeated_string_array = 2;
}
```

I also verified the following cluster packs/unpacks correctly.

image

# Reviewers
- Adding @bkeryan to focus on feedback of C++ changes.
- Adding @dixonjoel to focus on feedback of LV changes.
- Everyone else is FYI


src/well_known_messages.cc
```diff
@@ -0,0 +1,248 @@
...
+        //---------------------------------------------------------------------
+        std::shared_ptr<MessageMetadata> MetadataOwner::FindMetadata(const std::string& name)
+        {
+            return _owner->FindLocalMetadata(name);
+        }
+
+        //---------------------------------------------------------------------
+        //---------------------------------------------------------------------
+        template<typename TRepeatedType>
+        std::shared_ptr<MessageMetadata> Base2DArray<TRepeatedType>::GetMetadata(IMessageElementMetadataOwner* metadataOwner)
```
bkeryan, line 45: FYI: defining template methods in the .cpp file only works if the template is instantiated from the .cpp file. In this case, it happens to work because the same .cpp file defines a subclass for each template instance.

I guess this is fine, but if it breaks, the solution is to either move the code to the .h/.ipp file or add explicit instantiation to the .cpp file.

src/well_known_messages.cc
```diff
@@ -0,0 +1,248 @@
...
+            static const std::string messageUrl = "ni.protobuf.types.String2DArray";
+            return messageUrl;
+        }
+
+        //---------------------------------------------------------------------
+        //---------------------------------------------------------------------
+        void String2DArray::CopyArrayFromMessageToCluster(int rows, int columns, const std::shared_ptr<const LVMessageValue>& dataFieldValue, int8_t* start)
+        {
+            auto elementCount = rows * columns;
+            NumericArrayResize(GetTypeCodeForSize(sizeof(char*)), 2, start, elementCount);
```
bkeryan, line 216: nit: the array doesn't actually contain char pointers
```suggestion
            NumericArrayResize(GetTypeCodeForSize(sizeof(LStrHandle)), 2, start, elementCount);
```

src/well_known_messages.cc
```diff
@@ -0,0 +1,248 @@
...
+        void String2DArray::CopyArrayFromClusterToMessage(int totalElements, LV2DArrayHandle array, const std::shared_ptr<LVMessageValue>& dataFieldValue)
+        {
+            auto lvStr = (*array)->bytes<LStrHandle>();
+            auto dataValue = std::static_pointer_cast<LVRepeatedMessageValue<std::string>>(dataFieldValue);
+            for (int i = 0; i < totalElements; i++)
+            {
+                auto str = GetLVString(*lvStr);
+                dataValue->_value.Add(str);
+                lvStr += 1;
+            }
```
bkeryan, line 245: I think you can avoid some unnecessary copies here.

```suggestion
            auto dataValue = std::static_pointer_cast<LVRepeatedMessageValue<std::string>>(dataFieldValue);
            dataValue->_value.Reserve(totalElements);
            for (int i = 0; i < totalElements; i++)
            {
                auto str = GetLVString(*lvStr);
                dataValue->_value.AddAlreadyReserved(std::move(str));
                lvStr += 1;
            }
```

If `Add(std::move(str))` doesn't compile then I think our protobuf library is outdated. The docs in https://protobuf.dev/reference/cpp/api-docs/google.protobuf.repeated_field/ say `Add` takes `const Element&` but the code in https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/repeated_field.h takes `Element` by value in order to support move semantics.

If we are using an outdated protobuf then this might work:
`*dataValue->_value.AddAlreadyReserved() = std::move(str);`

or you can resign yourself to making unnecessary string copies. At least do the reserve, though.

bkeryan (changes requested):

src/well_known_messages.cc
jasonmreding: I moved the implementation to the header.

src/well_known_messages.cc
jasonmreding: Yeah, I originally had `void*` and then changed to `char*` with similar logic in `cluster_copier.cc`. I like use of `LStrHandle` better and have updated both locations.

src/well_known_messages.cc
jasonmreding: Seems to work fine. Thanks for the suggestion. I also updated similar code from `cluster_copier.cc`.

bkeryan (approved):

PR #419, opened 2025-04-02
Update server dll path to better serve the RTE use case out of a LV Source Distribution

SEbert-NI:
Update dll search to find dll in an adjacent location for a LV Source Distribribution Build running out of the RTE using this library within the source distribution


yash-ni:
Hi @SEbert-NI , would it be possible for you to add screenshots of the VI changes in the PR description? It would really help make the review process a bit easier. Thanks?

SEbert-NI:
image
image

labview source/gRPC lv Support/Server API/Server/Get Server DLL Path.vi
jasonmreding, line 1: I'm trying to make sure I understand the desired workflow. If you have a set of VIs which utilize grpc-labview (presumably some client VIs calling into some grpc service) and you want to execute them from TestStand using LV adapter configured to use LabVIEW Runtime-Engine rather than LabVIEW Development System, then you would:
1. Create source distribution buildspec in LV
1. Configure the buildspec to:
   - Not exclude any dependencies from vi.lib, instr.lib, or user.lib. 
   - Not exclude dependent packed libraries or shared libraries.
   - Embed compiled code into source files
   - Maybe apply prefix to avoid potential name conflicts if using multiple source distributions? Although I guess this probably won't help with share library dependencies that cannot be auto detected by the build.
   - Manually copy grpc library dlls to same location as build copies this dependent VI via pre/post build step.
1. Build
1. Point TS sequence at source distribution output of the build (or wherever you copy it to on disk)
1. Run TS sequence and everything now works

Is that an accurate description?

PR #421, opened 2025-04-08
fixed wrong get indicator label and added reverse 1d array to ctrl se…

pettaa123:
…arch

GetOneOf_Get_Accessor.vi search for wrong label Get_Indicator, instead of Get_Val as in OneOfTemplate.lvclass:GetTemplate.vi
This worked since Get_Val was always last in ctrl array.
Because of this I added a reverse 1d array to ctrl array, since Get_Val and Set_Val appear to be always last.

image

image

image


labview source/Client Server Support New/gRPC Scripting Tools/Oneof API/Accessor Helpers/GenOneOf_Get_Accessor.vi
jasonmreding, line 1: The `Controls[]` property returns the control references according to their tabbing order. By default, I believe the tabbing order is the same as the order the controls were dropped on the front panel. Given this, inserting a `Reverse 1D Array` node is just as arbitrary as reconfiguring the tabbing order so the control of interest is at tab stop index 1. Of the two options, my preference leans towards updating the tab order instead. 

If you want to go for extra credit, you can also update the search loop to pass out a found Boolean and then add a case structure that assigns an error if we can't find the control we are looking for. I typically just use `Error 5000.vi` for this sort of thing where it is more of an assert than anything else. These same comments would apply to `GenOneOf_Set_Accessor.vi` as well.

pettaa123:
I didn't know about the tabbing order, I learned that. Thanks

jasonmreding (approved):

jasonmreding:
@pettaa123 It appears one or all of the VIs from this PR were saved in LV 2020 rather than LV 2019.
https://github.com/ni/grpc-labview/actions/runs/14582170845/job/40977143664?pr=421#step:6:32

jasonmreding (changes requested):
Please resubmit VIs saved in LV 2019.

Issue #420, opened 2025-04-04
building daqmx client errors after a an hour


pettaa123:
Hi,

I am struggling to build daqmx client. If fails after an a few hours compiling with this error:

![Image](https://github.com/user-attachments/assets/8a45a31c-3f34-45f2-b1a3-8b5c2d48b0be)

Any suggestions?

[AB#3073418](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3073418)

pettaa123:
![Image](https://github.com/user-attachments/assets/cc6fea72-36ad-4f1e-ae91-c0e4e4e75422)

jasonmreding:
I am not sure how much time I will have to continue looking at this given the nature of the problem, but I was able to reproduce the failure you reported. In my case, it was ~4.5 hours into generation before failing, and it had generated 5,709 files before it failed. I don't know if it failed similarly for you or how deterministic the failure is.

From examining the stack trace that is reported, I can see the failure is coming from the `Get or Create Typedef.vi` shown below:

![Image](https://github.com/user-attachments/assets/ea663587-d681-4841-baf5-8ccb78aa124b)

Just from code inspection, it appears there might be a race condition of sorts between the top and bottom railroad tracks. The top railroad track is going to try to delete the typedef from the library if it already exists while the bottom railroad track is going to create a new typedef with the same file name (presumably stand alone and not part of any library just yet). Since there is no data flow dependency between the deletion of the old typedef and the creation of the new typedef, it is a race as to which will finish first. The delete in this case is really a removal of the item from the library it belongs to and not a delete from memory/disk. By removing it from its owning library, what I call the "link name" of the typedef is changed since it is no longer scoped to a library. This creates an opportunity for a name collision if there is already a typedef in memory in the global namespace with that same name. I'm pretty sure that is what error 2 is referring to in this case (not a memory full error). LabVIEW just does a horrible job of indicating what really went wrong in this case. 

To be honest, I'm having a hard time understanding why there would ever be an existing typedef for that name in the library such that we would need to remove it, but clearly it is happening. That assumes a clean generation from scratch and not trying to regenerate on top of an already generated client library. Perhaps if you are able to identify the typedef that is producing the failure, we can correlate back to the .proto file and figure out what might be going wrong. 

If you want to take a stab at debugging/fixing on your own, I would probably start by either:
[...]

PR #425, opened 2025-04-11
moved close reference out of case structure

pettaa123:
Related to #420 

Moved close reference out of case structure

image


labview source/Client Server Support New/gRPC Scripting Tools/Project API/Find Item in ProjectItem.vi
jasonmreding, line 1: Not sure what happened, but the VI source was saved as LV 2025 which will break backward compatibility with older versions. The project for the generator should be configured so source is saved as LV 2019 (as long as you are editing with LV 2024+). So as long as you edit/save the VI in the context of the project, it should automatically save in the correct version. You can validate that by looking in the upper right corner of the VI for the save version.

image

If you don't see a version number displayed, then it will save in the current version of the editor.

jasonmreding, line 1: From the PR description screenshot, the error in/out of the `Close Reference` is not wired. Please snake the error wire through it for consistency. It's unlikely that will ever generate an error. However, if it does, automatic error handling popups and what not can play havoc with the UI and automated tests.

jasonmreding, line 1: There are three VIs that are basically the same: `Find Item in ProjectItem.vi`, `Find Item in Library.vi`, and `Find Item in LVClass.vi`. The `Find Item in LVClass.vi` also has the same problem which would be good to fix at the same time as this.

**Optional:** 
The only reason there are three VIs is they have different refnum in/out types, and LV doesn't support polymorphic in/out types like this with a single VI unless the VI is part of a class where the terminals are configured appropriately. However, all of the refnum types are of type `ProjectItem` so we could centralize the logic using that interface/type and then create wrappers around that for class and library project item types. So the block diagram would basically be a single sub VI call to the centralized VI where the library/class item wire will coerce up to the project item type. You can then either downcast the out refnum type back to the original type or just leave it unwired and wire the refnum in/out controls directly to each other. Hopefully all of that makes sense. This would be my preferred solution to single source the code, but if you don't have time for that level of rework, I will accept the corrections to the existing code as is.

jasonmreding (changes requested):
Waiting for source version fix before approving CI tests to run.

labview source/Client Server Support New/gRPC Scripting Tools/Project API/Find Item in ProjectItem.vi
pettaa123: image

I also wired the ProjectItem property node error out, despite this was not consitent between the three.

jasonmreding (approved):

PR #426, opened 2025-04-11
Fixing reference leak when creating clients due to incorrect CleanupProc 

### LINKED ISSUE #379: Memory Reference Leak from Create client node
Running the create client node is showing some unknown memory reference leak in desktop execution trace toolkit
image

AB#2844161

jasonmreding:
# Description
Fixes #379 

This issue also causes LV to hang on exit if the user does not explicitly call `Close Client.vi` or aborts execution before it can be called. The shell disappears but the process remains in task manager forever unless manually terminated. Subsequent launches of LV will create a new process that may hang on exit as well creating more leaked processes over time.

The issue is caused because the `CreateClient` entry point registers a pointer using `grpc_labview::gPointerManager.RegisterPointer`. However, the LV cleanup proc it registers does not call `grpc_labview::gPointerManager.UnregisterPointer`. This leaks a shared pointer to a gRPC `Channel` which prevents ref counts from hitting zero until the process terminates and the dll is unloaded. However, by this time, all the threads for the process have terminated so those threads cannot react to ref counts finally hitting zero in order to cleanup other state. This ultimately causes 
`grpc_timer_manager_shutdown` to hang on shutdown as it synchronously blocks waiting for a conditional variable to reach the expected state when all grpc threads have been shutdown gracefully.

# Implementation / Design
**C++:**
- Cleanup proc now just calls `CloseClient` the same as the `Close Client.vi`.
- Added new exported entry point `CreateClient2` which allows caller to specify when the cleanup proc is run (when idle vs. after reset). `CreateClient` forwards calls to `CreateClient2` with a default cleanup behavior of cleanup when idle. I added this capability since [measurement-plugin-labview](https://github.com/ni/measurement-plugin-labview) uses a "client pool" at the G layer since there is no channel pool internal to the library and this fix effectively renders it useless since it creates clients from VIs that are called by reference. I considered adding a channel pool to default implementation of `LabVIEWgRPCClient`. However, I believe the grpc libraries used here still limit the maximum number of concurrent calls on a channel before serializing execution of calls (where as C# will automatically spin up new connections as needed to avoid this). See https://github.com/grpc/grpc/issues/21386 which has been open for a while now. Because of this, I didn't feel comfortable making this change without a way for the user to opt out, and I didn't want to introduce public APIs around that at the moment.

**LabVIEW:**
- Modified `Create Client.vi` to call new `CreateClient2` entry point and added a new recommended input for `auto cleanup` where the default value is TRUE.

image

# Testing
Manually tested that LV process now exits cleanly when closing LV after creating a client without explicitly calling `Close Client.vi`. I also verified via debugger that cleanup proc fires appropriately based on value of `auto cleanup` input.


labview source/gRPC lv Support/Client API/Create Client.vi
bkeryan, line 1: image

> If <b>auto cleanup</b> is FALSE, LabVIEW does not automatically destroy the client until you exit LabVIEW.

I think the **Abort Execution** button on the toolbar will also destroy the client.

bkeryan, line 1: image
block diagram diff, for reference

bkeryan (approved):
Approved with suggestion

labview source/gRPC lv Support/Client API/Create Client.vi
jasonmreding: Yes, it also performs cleanup when execution is aborted. I updated the documentation to indicate this.

PR #430, opened 2025-04-17
Improve the AnyBuilder support for enum types

chadallee:
Currently, passing MessageElementType=EnumValue will always crash.
We're constructing a `metadata` object in `ClusterDataCopier::AnyBuilderAddValue` that has no `_owner`, but then we pass that to `ClusterDataCopier::CopyEnumFromCluster`, which immediately tries to dereference metadata->_owner:

```cpp
    bool ClusterDataCopier::AnyBuilderAddValue(LVMessage& message, LVMessageMetadataType valueType, bool isRepeated, int protobufIndex, int8_t* value)
    {
        auto metadata = std::make_shared<MessageElementMetadata>(valueType, isRepeated, protobufIndex);
        switch (valueType)
        {
        // ...
        case LVMessageMetadataType::EnumValue:
            CopyEnumFromCluster(metadata, value, message);
            break;

// ...

    void ClusterDataCopier::CopyEnumFromCluster(const std::shared_ptr<const MessageElementMetadata> metadata, int8_t* start, LVMessage& message)
    {
        std::shared_ptr<EnumMetadata> enumMetadata = metadata->_owner->FindEnumMetadata(metadata->embeddedMessageName);
```   

So, first is a minor fix: throw `InvalidEnumValueException` instead of dereferencing null, which translates to a nice `-2003` error code in LabVIEW.

But the bigger fix is to better support that enum data type w/ the Any builder part of the API. I did that by adding a new exported function called `AnyBuilderAddValueFromMetadata` that takes in a metadata owner (i.e. the grpc client) and looks up the enum metadata from there.

Here's what it looks like on the LV side:
image

And I also added a test to the "deprecated" section side by side with the existing **SimpleAnyBuilder.vi** test.
This one tests the interoperability with the existing `AnyBuilderAddValue` and also tests the error case with an invalid enum value:

image

This test required making `Element to Enum Metadata.vi` public


labview source/gRPC lv Support/grpc-labview.lvproj
jasonmreding, line 1: Please update the project file to reference all of the VIs in this folder on disk.

tests/Deprecated/Protobuf/EnumAnyBuilderFromMetadata.vi
jasonmreding, line 1: Please add a reference to this VI in Tests.lvproj.

jasonmreding, line 1: You could also use `Cluster to gRPC LabVIEW Message.vi` to simplify the diagram a bit. I'm not sure if customers would use that API when packing/unpacking an `Any` or not. It's not currently on the palette. Either way, we could still use it for tests if it is convenient. You are welcome to accept/reject this suggestion as you see fit.

labview source/gRPC lv Support/grpc-labview.lvproj
jasonmreding, line 1: This is kind of a general topic for the overall PR, but I'm attaching to this file in order to get a threaded conversation.

When I first looked at the test as an example usage of the new API, I was a bit confused at first. In particular, I was thinking the message name supplied was for the enum itself and not the message which owns the enum field. If we went that route, then the diagram would look something like:

image

which only requires registering metadata for the enum and not the entire top level message which owns the enum field. If you want to unpack the message using the same serialization session, then you would still need to supply metadata for the entire message. However, I'm not sure how common that will be. In most cases, I kind of expect the code will either be packing or unpacking but not both. 

Another advantage of this approach is you only need to supply the definition of the enum type once rather than for each field in which the type is used. So rather than the protobuf index being used as a lookup into an existing message, it specifies the index for the field being written like it does for the existing `AnyBuilderAddValue.vim`, and the name of the enum message itself becomes the lookup.

For the above to work, the API would have to look something like:

image

which is spatially challenged for connector pane space. However, you could do this with the existing API VI by adding recommended terminals rather than introducing a new API. The downside of reusing the existing API is there isn't a good place to run an in/out terminal pair for the serialization session. I kind of wish the builder refnum itself represented a `MessageElementMetadataOwner` rather than a `LVMessage` which would help with some of this. I think that would be possible without breaking anything, but it also has its own pros and cons if you want to use the same serialization session across multiple builders. That con could probably be overcome by adding an optional input to `AnyBuilderBegin.vi` such that it can use an existing serialization session if wired and otherwise has its own serialization context so that you don't always have to manage two refnums through all of the APIs.

Anyway, I will likely have other comments for this PR, but I wanted to have a higher level discussion on the appropriate public API first. Presumably you are adding this because you have a need for it so you probably have a better perspective on expected usage of the pack/unpack APIs then I do at the moment (or at least a different perspective). My experience with the API is simply converting a cluster to an `Any` and vice versa and not packing a message one field at a time. I personally kind of like the idea of the AnyBuilder having a built in serialization context without necessarily having to provide the metadata for the owning message of a message field, but I also recognize it has a higher implementation cost.

chadallee:
After talking with @jasonmreding I've decided to abandon this PR.

The AnyBuilder API isn't designed to require metadata. So the alternative is to just call AnyBuilder with type=Int32 instead of type=Enum. The only thing we lose in that case is enum value validation (make sure the int value is valid for the given enum).

But in the AnyBuilder case, the caller would need to register that enum data anyway, so they already have access to all the legal values, so they can just validate enum values themselves.

Maybe we'll revisit this some day if we need to better support the "Message" types for stuff like waveforms & 2d arrays.

PR #431, opened 2025-04-21
Fix AnyBuilder(Type=Enum) crash

chadallee:
For AnyBuilder, just treat enum as int32.

When called from `AnyBuilderAddValue()`, `ClusterDataCopier::CopyEnumToCluster()` was crashing trying to dereference the null `_owner` element.

This change causes `AnyBuilderAddValue()` to not call `ClusterDataCopier::CopyEnumToCluster()` at all, but let's check pointers in `ClusterDataCopier::CopyEnumToCluster()` anyway.

See https://github.com/ni/grpc-labview/pull/430 for commentary on my original attempt to better support enums in AnyBuilder.

### DIFF
--- a/src/cluster_copier.cc
+++ b/src/cluster_copier.cc
@@ -222,7 +222,9 @@ namespace grpc_labview {
             CopyUInt64FromCluster(metadata, value, message);
             break;
         case LVMessageMetadataType::EnumValue:
-            CopyEnumFromCluster(metadata, value, message);
+            // Treat enum as an int32 for now.
+            // CopyEnumFromCluster() requires enum metadata and the AnyBuilder API doesn't allow users to provide that.
+            CopyInt32FromCluster(metadata, value, message);
             break;
         case LVMessageMetadataType::SInt32Value:
             CopySInt32FromCluster(metadata, value, message);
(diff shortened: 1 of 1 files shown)


src/cluster_copier.cc
jasonmreding, line 1: Is there anything from `UnpackedFields::GetField` that should be updated as well? It looks like that code already treats Int32 and enum values the same so probably not. From my read of the code, it looks like that method will always return -1 error anyway so maybe that code isn't ready for public use anyway? Just wanted to double check you don't also need serialize data in the opposite direction.

jasonmreding (approved):

src/cluster_copier.cc
chadallee: Yeah, from reading the code, I agree it seems to treat enums as int32.
I locally modified the `tests\Deprecated\Protobuf\TestUnpackFields.vi` to pack & unpack an enum and it worked fine.

I didn't see it crash, but passing a u8/u16 enum from LV is probably not safe.
But it's no less safe than passing a non-enum u8/u16 into AnyBuilder w/ type set to something bigger (int32/double/etc)

Seems fine.

PR #335, opened 2024-01-03
Client side communication performance improvements

CPattar-NI:
Optimizing the copy path between LV and protobuf​ by eliminating data copy into temporary structures. Instead of the previously used "values" map to store the message data temporarily during the message parsing phase and then walking through the map to fill in the LV structure, we copy the message values directly into LV structures.

PR #432, opened 2025-04-22
gRPC upgrade to v1.62.0

Sacheta-NI:
Upgraded the gRPC submodule from v1.44 to v1.62.0 to address multiple security vulnerabilities (CVE-2023-44487, CVE-2023-4785, CVE-2023-33953).

Temporarily disabled LVMessageEfficient due to a breaking change in the updated Protobuf library:
RepeatedField no longer supports the char type, which was used in the LVMessageEfficient class.
For now, the system falls back to using the LVMessage class. A permanent fix will be addressed separately.

Testing

1. Windows (32-bit and 64-bit):
   -   Automated tests executed using the New_ATS test suite.
   -   Manual LabVIEW client testing performed against New_ATS test cases.

2. Verified that LVMessageEfficient is properly disabled through feature toggle testing.

3. Linux testing has not been performed. During early validation only, we observed that several code generation VIs use hardcoded Windows-style paths instead of building paths dynamically with the "Build Path" node. These hardcoded paths are incompatible with Linux It must not be working even currently. Path corrections and Linux validation will be addressed separately.


src/well_known_messages.cc
```diff
@@ -128,12 +128,12 @@ namespace grpc_labview
...
         {
             auto lvStr = (*array)->bytes<LStrHandle>();
-            auto dataValue = std::static_pointer_cast<LVRepeatedMessageValue<std::string>>(dataFieldValue);
+            auto dataValue = std::static_pointer_cast<LVRepeatedStringMessageValue>(dataFieldValue);
             dataValue->_value.Reserve(totalElements);
             for (int i = 0; i < totalElements; i++)
             {
                 auto str = GetLVString(*lvStr);
-                dataValue->_value.AddAlreadyReserved(std::move(str));
+                dataValue->_value.Add(std::move(str));
```
jasonmreding, line 136: Is there a reason this cannot still use `AddAlreadyReserved`?

yash-ni: @jasonmreding Since _value is of `RepeatedPtrField` type, it doesn’t support an `AddAlreadyReserved` method.

src/message_value.h
```diff
@@ -137,6 +138,20 @@ namespace grpc_labview
         google::protobuf::uint8* Serialize(google::protobuf::uint8* target, google::protobuf::io::EpsCopyOutputStream* stream) const override;
     };
 
+    class LVRepeatedStringMessageValue : public LVMessageValue
```
jasonmreding, line 141: Can you provide some context on why this new message type was needed? From inspection, it appears it was just to optimize out the `_cachedSize` field? If that is the case, I am not opposed to this. However, the class hierarchy is a bit of a mess as there are other repeated message types that also do not utilize this field so it seems strange to optimize one of them and not all of them. If this is the goal, maybe this change would be better as a separate PR where all of the types are optimized at once? There is a lot of duplicated code throughout that seems like it could be cleaned up with some small refactoring to the inheritance hierarchy.

src/lv_message_efficient.cc
```diff
@@ -172,9 +172,9 @@ namespace grpc_labview
             auto repeatedMessageValuesIt = _repeatedMessageValuesMap.find(fieldInfo.fieldName);
             if (repeatedMessageValuesIt == _repeatedMessageValuesMap.end())
             {
-                auto m_val = std::make_shared<RepeatedMessageValue>(fieldInfo, google::protobuf::RepeatedField<char>());
+                auto m_val = std::make_shared<RepeatedMessageValue>(fieldInfo, google::protobuf::RepeatedPtrField<google::protobuf::Message>());
                 repeatedMessageValuesIt = _repeatedMessageValuesMap.emplace(fieldInfo.fieldName, m_val).first;
-                repeatedMessageValuesIt->second.get()->_buffer.Resize(arraySize, _fillData);
```
jasonmreding, line 177: `arraySize` is no longer used

src/well_known_messages.h
```diff
@@ -126,7 +126,18 @@ namespace grpc_labview
                 arrayMessage->_values.emplace(_rowsIndex, rowsValue);
                 auto columnsValue = std::make_shared<LVVariableMessageValue<int>>(_columnsIndex, columns);
                 arrayMessage->_values.emplace(_columnsIndex, columnsValue);
-                auto dataValue = std::make_shared<LVRepeatedMessageValue<TRepeatedType>>(_dataIndex);
+                //auto dataValue = std::make_shared<LVRepeatedMessageValue<TRepeatedType>>(_dataIndex);
```
jasonmreding, line 129: Delete commented out line of code.

jasonmreding:
What testing has been done as part of this change?

src/lv_message_efficient.h
```diff
@@ -29,19 +29,19 @@ namespace grpc_labview
     protected:
         struct RepeatedMessageValue {
             const MessageElementMetadata& _fieldInfo;
-            google::protobuf::RepeatedField<char> _buffer;
+            google::protobuf::RepeatedPtrField<google::protobuf::Message> _buffer;
             uint64_t _numElements = 0;
 
-            RepeatedMessageValue(const MessageElementMetadata& fieldInfo, google::protobuf::RepeatedField<char> buffer) :
+            RepeatedMessageValue(const MessageElementMetadata& fieldInfo, google::protobuf::RepeatedPtrField<google::protobuf::Message> buffer) :
```
jasonmreding, line 35: Presumably this will have worse performance? At least for clusters/messages where all the data is "flat", you will end up with an allocation per element rather than one allocation for the entire repeated field/array? I'm not necessarily opposed to this sort of change, but if there isn't a clear performance goal in mind, then I would prefer not to maintain both `LVMessage` and `LVMessageEfficient` classes which has already proven itself to be a maintenance burden. Unfortunately, I don't have any context for what motivated the introduction of `LVMessageEfficient` or why its use wasn't promoted more broadly so we don't have to maintain multiple serialization code paths.

.github/workflows/windows_x64_build.yml
```diff
@@ -14,13 +14,17 @@ jobs:
     # well on Windows or Mac.  You can convert this to a matrix build if you need
     # cross-platform coverage.
     # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
-    runs-on: windows-2019
+    runs-on: windows-latest
```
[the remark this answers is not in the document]
Sacheta-NI: Yes, right. Have updated in both the ymls: x64 and x86

src/well_known_messages.h
Sacheta-NI: Yes, Done.

src/lv_message_efficient.h
jasonmreding: I don't see how any of this is going to work. If you look at `LVMessageEfficient::PostInteralParseAction` you will see we eventually perform a block `memcpy` from this buffer into the memory allocated by LV to store the array of clusters. With this change, you are no longer dealing with a byte buffer and are instead dealing with a buffer of pointer sized data. That can work as long as the cluster size is less than a pointer size. Otherwise, it seems like you are going to overrun the buffer.

Unless these source changes are needed to upgrade the grpc dependency, my recommendation would to revert them and keep this PR to just the update in the dependency. You can then perform refactoring/optimizations in later PRs as desired. This will allow us to better isolate the cause of any regression failures.

jasonmreding (changes requested):
Waiting for changes to be reverted or confirmation via testing that serialization from a LV client for an array of messages works.

src/message_value.h
Sacheta-NI: Hey @jasonmreding ,
This class wasn't created for optimization purposes, but rather to meet the requirements of the upgrade. Previously, we were using the LVRepeatedMessageValue template class for strings as well, which internally relies on google::protobuf::RepeatedField.

However, after the upgrade, we started encountering compilation errors because google::protobuf::RepeatedField now only supports primitive types like int32_t, int64_t, uint32_t, uint64_t, float, double, and bool. It no longer supports strings.

As a result, we had to switch to google::protobuf::RepeatedPtrField for handling strings, which led us to introduce a new class specifically designed for handling string values.

src/lv_message_efficient.h
Sacheta-NI: Hey @jasonmreding ,
You're right — it looks like the changes in this file won't work as expected. These updates were originally made as part of the upgrade effort, since RepeatedField no longer supports char, and unfortunately it doesn't support int8_t either.

I had assumed that LVMessageEfficient was still disabled, but it seems it’s been enabled recently. Interestingly, the New_ATS tests didn’t catch this issue — they all passed. I even ran the same test cases manually using the LV client, and they passed there too.

We should definitely add a specific test case to cover this scenario in our suite.

I’ll take a look and follow up with an update.

jasonmreding: I did not realize the grpc upgrade changed behavior in regards to what types are allowed with a repeated field. Probably the simplest thing to do then is to just switch to using a `std::vector`. Maybe we should have just used that to begin with since that is more or less how we were using the field anyway.

Yes, our automated testing around this is pretty lacking. I believe most of the testing occurs via a python client communicating with a LV service and that there is little to no testing of the LV client itself. IIRC, this message type is only used on the client side of a unary call. Again, I don't know what motivated the introduction of this optimized message type. I just know it was enabled via the default feature flags and had a lot of bugs which required fixing.

Sacheta-NI: Hey @jasonmreding 
Addressing and validating the changes in LVMessageEfficient will require some additional time. For this release, we’re considering keeping the current changes to allow compilation to proceed, with LVMessageEfficient temporarily hardcoded to off. We’ll file a bug to track this and plan to fix and test it properly in a future update.
Please let us know if you have any concerns.
cc: @CPattar-NI @yash-ni

jasonmreding: Again, I don't know what motivated the introduction of this optimized code path so I cannot speak to potential impact of degraded performance. I personally don't have any objections to disabling the code path as long as you are certain it is really disabled via the feature flags. In particular I would want to ensure:
- Things continue to work on systems with existing ini files where the feature flag is enabled.
- The default generation of the ini file doesn't activate the broken code path. I think there is some duplication between the C code and LV code so disabling the default value in the C code isn't always sufficient. I believe there is also some LV code which will trigger generation of the ini file which doesn't pay attention to default values defined in the C code. I don't know if that means you will have to hard code the conditional branch rather than relying on the feature flag state.
- There is some commitment to fixing the broken code path in a reasonable time frame or we simply get rid of that code path and utilize the standard `LVMessage` for everything.

Is the motivation for updating to the latest grpc source to address known CVEs that have since been fixed? Just trying to better understand the trade offs here.

Sacheta-NI: Hey @jasonmreding ,
I found the PR where LVMessageEfficient was introduced: https://github.com/ni/grpc-labview/pull/335 .It was focused on client-side optimizations, and from my discussion with @CPattar-NI, I recall it brought about an 8x improvement in performance wrt time in case of arrays.

Have discussed with him, it's fine if we temporarily disable it, fix it later and release post proper testing.

As for disabling it, I've set the token's default value to false in feature_toggles.h and overridden it in the token map after populating it from the ini file (changes in feature_toggles.cc). I've tested it by setting it to true in the feature_config.ini file, and it's working as expected.

Regarding the timeline for the fix, I'm not sure at the moment since we're currently low on bandwidth.

src/lv_message_efficient.cc
Sacheta-NI: We will be disabling this class for now, make a proper fix and then enable.

jasonmreding:
> Linux testing has not been performed due to hardcoded paths in VIs which are compatible with windows, not Linux. It must not working even currently. Path corrections and Linux validation will be addressed separately.

Can you elaborate on this a bit? Is this an issue with the lookup path for the grpc library on Linux?

jasonmreding (approved):

yash-ni (approved):

src/feature_toggles.cc
```diff
@@ -46,6 +46,9 @@ namespace grpc_labview {
         }
 
         configFile.close();
+        //TODO: remove this post fixing LVMessageEfficient to let enable feature
```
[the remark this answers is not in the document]
Sacheta-NI: Done. Have added issue number in comment. Here's the link to issue: https://github.com/ni/grpc-labview/issues/433

CPattar-NI (approved):

Issue #195, opened 2022-11-15
Proper Error Code Handing


carlschwabni:
At the moment, the gRPC drivers are presenting error codes in the range of -1016 to -1001 and -5 to -1, and an error codes files is not being deployed with the VIPM package.

Ideally:
- The error codes would use a dedicated, valid error range. This could be done by shifting all error codes coming from the C call by a constant value (e.g. -1016 becomes -821016, with a different shift applied to -5 to -1). Note that "-820000" was randomly chosen for demonstrative purposes, and isn't what you'd really want to use.
- Deploy an [error code file](https://www.ni.com/docs/en-US/bundle/labview/page/lvhowto/def_custom_error_text.html) with the package that defines the errors.

Special considerations:
- How will this impact existing user base/how big is user base?  If users are already catching specific errors, changing error codes could be problematic, but it'd still be best to deploy an error codes file.

bkeryan:
https://github.com/ni/grpc-labview/pull/436 highlights that this is still an issue: the code is full of `return -1` and `return -2`

PR #436, opened 2025-05-06
Changes to Better Support Asynchronous Server Shutdown

jasonmreding:
# Description
This PR better supports server shutdown when there are outstanding RPCs at the time of shutdown.

# Implementation / Design
- All exported APIs now first verify that the registered pointer passed in (gRPCid) is valid before using it and return an error if it is not valid. This was done predominantly for the pack/unpack APIs since they can reference the server object as the metadata owner. However, it was applied to all exported APIs for consistency.
- Updated `SetResponseData` and `CloseServerEvent` so they first check to see if the call has been canceled before trying to call into the underlying infrastructure to complete the call. This is necessary because the call object holds onto pointers to objects owned by the server, but the call lifetime is not tied to the server lifetime. This results in accessing a deleted pointer unless the call first checks for cancellation. I briefly considered changing the pointers to smart pointers instead as a more holistic solution. However, that would result in the server (and other `IMessageElementMetadataOwner` objects) holding onto message metadata which contained a smart pointer back to itself which didn't seem ideal.

# Testing
I manually tested using a service that arbitrarily added a ten second wait to the RPC before returning. I then shutdown the server after a client unary call was in progress but before it completed.

### DIFF
--- a/src/any_support.cc
+++ b/src/any_support.cc
@@ -139,6 +139,7 @@ LIBRARY_EXPORT int32_t TryUnpackFromAny(grpc_labview::gRPCid* id, grpc_labview::
     {
         return -2;
     }
+
     if (grpc_labview::GetLVString(anyCluster->TypeUrl) != messageType)
     {
         return -1;
@@ -157,6 +158,11 @@ LIBRARY_EXPORT int32_t IsAnyOfType(grpc_labview::gRPCid* id, grpc_labview::AnyCl
     }
 
     auto metadata = metadataOwner->FindMetadata(messageType);
+    if (metadata == nullptr)
+    {
+        return -2;
+    }
+
     if (grpc_labview::GetLVString(anyCluster->TypeUrl) != metadata->typeUrl)
     {
         return -1;
--- a/src/any_support.cc
+++ b/src/any_support.cc
@@ -31,6 +31,11 @@ LIBRARY_EXPORT int32_t FreeSerializationSession(grpc_labview::gRPCid* sessionId)
 LIBRARY_EXPORT int32_t PackToBuffer(grpc_labview::gRPCid* id, const char* messageType, int8_t* cluster, grpc_labview::LV1DArrayHandle* lvBuffer)
 {
     auto metadataOwner = id->CastTo<grpc_labview::IMessageElementMetadataOwner>();
+    if (!metadataOwner)
+    {
+        return -1;
+    }
+
     auto metadata = metadataOwner->FindMetadata(messageType);
     if (metadata == nullptr)
     {
@@ -63,6 +68,11 @@ LIBRARY_EXPORT int32_t PackToBuffer(grpc_labview::gRPCid* id, const char* messag
 LIBRARY_EXPORT int32_t PackToAny(grpc_labview::gRPCid* id, const char* messageType, int8_t* cluster, grpc_labview::AnyCluster* anyCluster)
 {
     auto metadataOwner = id->CastTo<grpc_labview::IMessageElementMetadataOwner>();
+    if (!metadataOwner)
+    {
+        return -1;
+    }
+
     auto metadata = metadataOwner->FindMetadata(messageType);
     if (metadata == nullptr)
     {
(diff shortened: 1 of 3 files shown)


src/any_support.cc
```diff
@@ -31,6 +31,11 @@ LIBRARY_EXPORT int32_t FreeSerializationSession(grpc_labview::gRPCid* sessionId)
 LIBRARY_EXPORT int32_t PackToBuffer(grpc_labview::gRPCid* id, const char* messageType, int8_t* cluster, grpc_labview::LV1DArrayHandle* lvBuffer)
 {
     auto metadataOwner = id->CastTo<grpc_labview::IMessageElementMetadataOwner>();
+    if (!metadataOwner)
+    {
+        return -1;
```
bkeryan, line 36: Error code range is covered by https://github.com/ni/grpc-labview/issues/195

src/any_support.cc
```diff
@@ -63,6 +68,11 @@ LIBRARY_EXPORT int32_t PackToBuffer(grpc_labview::gRPCid* id, const char* messag
 LIBRARY_EXPORT int32_t PackToAny(grpc_labview::gRPCid* id, const char* messageType, int8_t* cluster, grpc_labview::AnyCluster* anyCluster)
 {
     auto metadataOwner = id->CastTo<grpc_labview::IMessageElementMetadataOwner>();
+    if (!metadataOwner)
+    {
+        return -1;
+    }
```
bkeryan, line 74: This codebase ought to make better use of C++ exceptions. Instead of duplicating these checks in each exported function, the checks could be single-sourced:
- Create a C++ exception class for an exception with an error code.
- Change `CastTo` to throw an exception with the appropriate error code when it fails
- Add a `TryCastTo` method for use cases that require casting to null, and update callers to use it when appropriate.
- Add an exception dispatcher method like
  ```cpp
  int32_t translateException() {
      try { 
          throw; 
      } catch (const MyExceptionWithErrorCode& e) {
          return e.code;
      } catch (const std::exception& e) {
          logException(e);
          return SOME_ERROR_CODE;
      }
  }
  ```
- Add a try/catch block to each exported function that uses the exception dispatcher:
  ```cpp
  try {
      ...
  } catch (const std::exception&) {
      return translateException();
  }
  ```
  It is possible to wrap this in a macro, but not necessary, since most of the logic is single sourced in the exception dispatcher function.

It already does some of this, but only for invalid enum values.

This is outside the scope of this PR, so I'm going to convert this comment to an issue.

src/any_support.cc
```diff
@@ -144,21 +167,26 @@ LIBRARY_EXPORT int32_t IsAnyOfType(grpc_labview::gRPCid* id, grpc_labview::AnyCl
 //---------------------------------------------------------------------
 //---------------------------------------------------------------------
 LIBRARY_EXPORT int32_t AnyBuilderBegin(grpc_labview::gRPCid** builderId)
-{   
+{
     grpc_labview::InitCallbacks();
 
     auto metadata = std::make_shared<grpc_labview::MessageMetadata>();
     auto rootMessage = new grpc_labview::LVMessage(metadata);
```
bkeryan, line 174: This codebase is letting C++ exceptions escape from exported C functions.

I'm going to convert this to an issue.

src/any_support.cc
```diff
@@ -130,10 +149,14 @@ LIBRARY_EXPORT int32_t TryUnpackFromAny(grpc_labview::gRPCid* id, grpc_labview::
...
+{
     auto metadataOwner = id->CastTo<grpc_labview::IMessageElementMetadataOwner>();
-    auto metadata = metadataOwner->FindMetadata(messageType);
+    if (!metadataOwner)
+    {
+        return -1;
+    }
 
+    auto metadata = metadataOwner->FindMetadata(messageType);
     if (grpc_labview::GetLVString(anyCluster->TypeUrl) != metadata->typeUrl)
```
bkeryan, line 160: This function should check `metadata`.

I noticed this because of the "return -1, return -2" vs. "return -1, return -1"

bkeryan (approved):
Approved with suggestions:
- Add one more metadata check

I will convert a couple of these comments into issues.

src/any_support.cc
bkeryan: https://github.com/ni/grpc-labview/issues/437

src/any_support.cc
bkeryan: https://github.com/ni/grpc-labview/issues/438

bkeryan:
> I briefly considered changing the pointers to smart pointers instead as a more holistic solution. However, that would result in the server (and other IMessageElementMetadataOwner objects) holding onto message metadata which contained a smart pointer back to itself which didn't seem ideal.

@jasonmreding If you did need to do something like this, you could use `std::weak_ptr` to break the reference cycle.

src/any_support.cc
jasonmreding: I added a null check here as well. The error codes here don't really matter as the return type at the VI layer API just converts this to a Boolean for success or failure.

bkeryan (approved):

Issue #437, opened 2025-05-06
grpc-labview ought to make better use of C++ exceptions. Instead of duplicating these checks in each exported function, the checks could be single-sourced:


bkeryan:
The grpc-labview C++ codebase has a lot of duplicated error handling, and https://github.com/ni/grpc-labview/pull/436 adds more.

Instead of duplicating checks for failed casts, etc. in each exported function, the checks could be single-sourced:
- Create a C++ exception class for an exception with an error code.
- Change `CastTo` to throw an exception with the appropriate error code when it fails
- Add a `TryCastTo` method for use cases that require casting to null, and update callers to use it when appropriate.
- Add an exception dispatcher method like
  ```cpp
  int32_t translateException() {
      try { 
          throw; 
      } catch (const MyExceptionWithErrorCode& e) {
          return e.code;
      } catch (const std::exception& e) {
          logException(e);
          return SOME_ERROR_CODE;
      }
  }
  ```
- Add a try/catch block to each exported function that uses the exception dispatcher:
  ```cpp
  try {
      ...
  } catch (const std::exception&) {
      return translateException();
  }
  ```
  It is possible to wrap this in a macro, but not necessary, since most of the logic is single sourced in the exception dispatcher function.

The code already does some of this, but only for invalid enum values.

_Originally posted by @bkeryan in https://github.com/ni/grpc-labview/pull/436#discussion_r2076039495_

bkeryan:
The feature toggle code would also benefit from being able to use C++ exceptions to report errors. Currently it ignores errors from 
opening `std::ifstream`, `GetModuleHandle`, etc. It makes sense to ignore file-not-found when using the default feature toggle file path, but other errors ought to be reported to the caller.

PR #443, opened 2025-05-12
Feature Toggle INI Efficiency Improvements

kt-jplotzke:
This pull request implements some efficiency improvements for the use of 'Feature Toggles'.  Specifically, three main changes are implemented:
- The `feature_config.ini` file is automatically read by the DLL when first needed (no longer called from LabVIEW)
- The `feature_config.ini` file is no longer created automatically
- The `feature_config.ini` file is not read by LabVIEW on every client call.

Previously, the `Wait on Occurrence Wrapper.vi` function read (and subsequently wrote, if the INI file did not exist) the `feature_config.ini` file.  `Wait on Occurrence Wrapper.vi` is called by `Client Unary Call.vim`, `Client Read from Stream.vim`, and `Client Complete Client Stream.vim`, so the feature INI file was being read essentially every time a gRPC client call was being made.  

I see two issues with this, in my opinion:
1. The feature toggle defaults were kept in two places -- Once in [`feature_toggles.h`](https://github.com/ni/grpc-labview/blob/master/src/feature_toggles.h) and another in `Wait on Occurrence Wrapper.vi`. 
2. The feature toggle INI values should only be used when _overriding_ a default value used by grpc-labview.  If the intent is to use all default "features", then no INI file should be required or created.  Creating a "default" version of the INI file also can give issues during updates -- If a feature default value ever changes in the future, automatically creating an INI file and writing a default value now can cause issues when that default value changes later.

As a result, this pull request has the following changes:

1. I've added code to read/parse the `feature_config.ini` file on the instantiation of the FeatureConfig object.  Features are now automatically loaded by the DLL and therefore the function `readIniFile` has been removed in `Get Server DLL Path.vi` 

2. I've added a new exported function to the DLL `IsFeatureEnabled`, which takes in feature name and returns if that feature is enabled (this is just a wrapper of `FeatureConfig::IsFeatureEnabled`).
```
LIBRARY_EXPORT int32_t IsFeatureEnabled(const char* featureName, uint8_t* featureEnabled)
{
    *featureEnabled = grpc_labview::FeatureConfig::getInstance().IsFeatureEnabled(featureName);
    return 0;
}
```

3. I've added a new LabVIEW function `Is Feature Enabled.vi`, which calls the DLL to fetch if a feature is enabled. 

4. `Wait on Occurrence.vi` has been updated to remove the `ms timeout` input and always leave the occurrence timeout as -1 (infinite) msec.  Previously, the occurrence was used to determine if a gRPC call timeout had occurred and subsequently threw an error if the occurrence timeout was exceeded.  As all timeouts are handled by the DLL (and errors are thrown appropriately from the CompleteClientCall functions), this `ms timeout` input was no longer used (and shouldn't be used).  As a result, this input was removed to avoid confusion.
Subsequently, the `Wait on Occurrence Wrapper.vi` functionality has been incorporated into `Wait on Occurrence.vi` and the former VI deleted.
image

I've built and tested the updates for both Win64 and LinuxRT and all look good.


labview source/gRPC lv Support/Shared/Wait On Occurrence Wrapper.vi
jasonmreding, line 1: I don't think we should be exposing the `ms timeout` input terminal. In fact, I think we should rip it all the way out as it is never wired or used. The underlying `Wait on Occurrence.vi` tries to set error -1004 based on a timeout which indicates grpc deadline exceeded status code which is just wrong. Even if we used something other an infinite timeout, the occurrence itself should not be used to indicate deadline errors. That status should be funneled out as part of the grpc call itself. I think once you rip all of that out, there probably isn't a lot left and you could just collapse the two VIs into a single VI. 

We should also mark this VI as private to the library.

jasonmreding (changes requested):
Thanks for posting this PR! I like the spirit of what it accomplishes. I left some comments for a few other things I would like to see cleaned up before submitting.

labview source/gRPC lv Support/Shared/Get Feature Toggle Enabled.vi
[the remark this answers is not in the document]
kt-jplotzke: During my initial testing, caching the result saved ~10 usec per client call, where a full roundtrip unary call on localhost is ~140 usec on NILRT.

I'll admit that my primary use case is using grpc-labview in a large NILRT application where I'm trying to reduce the overhead time as much as feasible.  Thus, 5-10% execution time per client call would matter to me, whereas running on Windows you'd never see this savings.

That said, I totally see the desire for having a path for an automated test and how this disrupts that.  I'll do a few more tests and give some thought if there's a compromise.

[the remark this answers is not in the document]
jasonmreding: That's good context. I don't have a NILRT to test with so I'm a bit surprised that going through the grpc stack as well as our marshalling code from this library would be that fast. Conversely, I'm a bit surprised that such a simple function call would have that much overhead when compared to what is going on with the rest of the unary call. There is also some configuration in regards to the error checking applied to the call library node you could play around with. I'm not certain how much overhead it may eliminate from the call.

image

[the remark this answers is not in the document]
bkeryan: > During my initial testing, caching the result saved ~10 usec per client call, where a full roundtrip unary call on localhost is ~140 usec on NILRT.

What was the baseline for this comparison? Calling the LabVIEW INI VIs or calling the DLL?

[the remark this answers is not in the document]
kt-jplotzke: Calling the DLL.  Removing the LV INI VIs saved ~20-30 usec per call.

Once I finish implementing the changes discussed here, I'll redo my measurements and re-confirm baseline vs DLL vs LV cache.

src/grpc_interop.cc
```diff
@@ -134,6 +134,14 @@ LIBRARY_EXPORT void readIniFile(const char* filePath)
     grpc_labview::FeatureConfig::getInstance().readConfigFromFile(filePath);
 }
 
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+LIBRARY_EXPORT int32_t getFeatureEnabled(const char* featureName, uint8_t* featureEnabled)
```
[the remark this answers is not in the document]
kt-jplotzke: I made several updates to `feature_toggles.cc` for better useability based on what makes sense to me.

- Features are now automatically loaded from the default file (`feature_config.ini`) on instantiation of the `FeatureConfig` object on first call to `FeatureConfig::getInstance()`.  Calling `readIniFile` is removed from the LV `Get Server DLL Path.vi`.
- A new exported function `ReloadFeaturesFromFile` allows for re-initializing the features from a given file.  My intent of this function is to be used for automated tests later on.  Because of this, I have not created a LV method for this function; I would not expect that this function would be used during normal use.
- `ReloadFeaturesFromFile` takes in a string filepath.  If the string is empty, it will reload the default `feature_config.ini` file.  If a filename only is given, it will build a path in the same folder as the DLL/shared library.  If an absolute path is given, it will use that specific file.  (again, just planning for automated testing in the future)
- Note that I kept `ReloadFeaturesFromFile` to not return an error return code if the given file is not found (this is a void function right now).  This was to keep with the previous functionality where the `readIniFile` was also void so that if a `feature_config.ini` file was present or not present the LV code acted the same.  I did add a "feature" named `featureFileFound` which can be queried to tell if features were loaded from file or if just the default features are being used.  If you would rather have this function return an error code, I can easily update.

I've tested both the default functionality and `ReloadFeaturesFromFile` on both Windows and LinuxRT and both look to handle paths and function correctly.

labview source/gRPC lv Support/Shared/Get Feature Toggle Enabled.vi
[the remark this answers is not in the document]
kt-jplotzke: - [x] Feature name input terminal is now required 
- [X] Updated all FP controls/indicators to use lower case naming
- [X] Renamed function to `Is Feature Enabled.vi`
- [X] Updated to private scope

labview source/gRPC lv Support/Shared/Wait On Occurrence Wrapper.vi
kt-jplotzke: I ripped out the `ms timeout` input to `Wait on Occurrence.vi` and then subsequently deleted `Wait on Occurrence Wrapper.vi` and combined all functionality into one VI.  I removed throwing error -1004 as the occurrence is now set to always wait forever.  Updated linking to `Wait on Occurrence.vi` in `Client Unary Call.vim`, `Client Read from Stream.vim`, and `Client Complete Client Stream.vim`

image

src/path_support.cc
```diff
@@ -28,7 +28,7 @@ namespace grpc_labview
         {
             char path[MAX_PATH];
             if (GetModuleFileNameA(hModule, path, MAX_PATH)) {
```
bkeryan: LabVIEW 2025 Q3 will be long-path aware: https://forums.ni.com/t5/LabVIEW-Public-Beta-Program-in/New-Feature-Support-for-Windows-Long-Paths/td-p/4434839

This code is using a fixed-size path buffer of size `MAX_PATH`, so it will truncate the path to 260 characters.

Consider making this code long-path aware:
```suggestion
            std::string path(MAX_PATH);
            DWORD result = GetModuleFileNameA(hModule, path.data(), path.size());
            while (result >= path.size() && GetLastError() == ERROR_INSUFFICIENT_BUFFER) {
                path.resize(path.size() * 2);
                result = GetModuleFileNameA(hModule, path.data(), path.size());
            }
            path.resize(result); // remove padding at end
            if (result) {
```

src/path_support.cc
```diff
@@ -37,24 +37,11 @@ namespace grpc_labview
         if (dladdr(reinterpret_cast<void*>(&GetFolderContainingDLL), &dl_info) && dl_info.dli_fname) {
             char realPath[PATH_MAX];
             if (realpath(dl_info.dli_fname, realPath)) {
-                fullPath = StripPath(std::string(realPath));
+                fullPath = std::filesystem::path(realPath).parent_path();
             }
         }
```
bkeryan: FYI: `realpath` is not long path aware/safe, but I don't have a suggestion here.

https://www.man7.org/linux/man-pages/man3/realpath.3.html

> ## BUGS         [top](https://www.man7.org/linux/man-pages/man3/realpath.3.html#top_of_page)
>
> The POSIX.1-2001 standard version of this function is broken by
> design, since it is impossible to determine a suitable size for
> the output buffer, resolved_path.  According to POSIX.1-2001 a
> buffer of size PATH_MAX suffices, but PATH_MAX need not be a
> defined constant, and may have to be obtained using [pathconf(3)](https://www.man7.org/linux/man-pages/man3/pathconf.3.html).
> And asking [pathconf(3)](https://www.man7.org/linux/man-pages/man3/pathconf.3.html) does not really help, since, on the one
> hand POSIX warns that the result of [pathconf(3)](https://www.man7.org/linux/man-pages/man3/pathconf.3.html) may be huge and
> unsuitable for mallocing memory, and on the other hand [pathconf(3)](https://www.man7.org/linux/man-pages/man3/pathconf.3.html)
> may return -1 to signify that PATH_MAX is not bounded.  The
> resolved_path == NULL feature, not standardized in POSIX.1-2001,
> but standardized in POSIX.1-2008, allows this design problem to be
> avoided.

kt-jplotzke: I think that `realpath` __is__ long path aware.

From the man page:
> The resulting pathname is stored as a null-terminated string, up to a maximum of PATH_MAX bytes

The default PATH_MAX on linux is 4096 bytes. 
Also confirmed on NILRT 24.8:

```
admin@xxx ~ $ getconf PATH_MAX /
4096
```

src/path_support.h
jasonmreding, line 1: Please convert tabs to spaces in this file.

src/feature_toggles.h
```diff
@@ -1,32 +1,46 @@
...
     class FeatureConfig {
+    public:
+        // Singleton instance
+        static FeatureConfig& getInstance() {
+            static FeatureConfig instance;
+            return instance;
+        }
+
+        // Function to reload all features from file and re-initialize the feature map
+        void ReloadFeaturesFromFile(const std::string& filePath);
```
jasonmreding, line 16: I would prefer not to add or expose this method. If we want to dynamically control feature toggles at run time, then I think we should simply add APIs like `EnableFeature`/`DisableFeature` or something similar. Having to go through the file system to change the state of a toggle does not seem ideal. That means removing it from `ExportedFunctionList.json` as well.

src/feature_toggles.h
```diff
@@ -1,32 +1,46 @@
...
+        const std::string defaultConfigFileName = "feature_config.ini";
 
-        // Constructor to initialize with default values
-        FeatureConfig() {
+        // This stores the default feature configuration.
+        // During ReloadFeaturesFromFile(), this configuration is always applied first prior to reading the
+        // features configuration file
+        void ApplyDefaultFeatures() {
             featureFlags["gRPC"] = true; // Enable gRPC by default as an example, this will never be overridden by config file
+            featureFlags["featureFileFound"] = false;  // Used to indicate if the feature file was found/used during initialization
```
jasonmreding, line 30: While you are here, I would recommend just deleting the "gRPC" flag as it no longer serves a purpose. I think it was added so that an ini file would always get generated with at least one flag that could be used as a template for adding other flags.

jasonmreding (approved):

src/path_support.cc
bkeryan: I don't think that's necessary until LabVIEW supports passing a Unicode string to the call library function node. Until then, the path to the DLL must be representable in the current ANSI code page.

src/path_support.cc
bkeryan: Ok, 4096 bytes is large enough that I'm not going to worry about it.

src/feature_toggles.h
kt-jplotzke: My only thought was there could be automated test cases which have different feature_config.ini files created and then these are re-loaded for each test case.

Are you opposed to keeping the function and just removing the exporting it?  That way, if/when test cases are defined the structure is in the DLL if needed at that time?

jasonmreding: I can live with that. However, my expectation is that this would likely never be used by automated tests. Instead, individual test cases would be responsible for updating the feature flags as appropriate as part of the test setup. However, maybe that's just my bias from how I've seen feature flags incorporated into testing in the past. Doing that is generally easier in my experience than requiring tests to manage deployment of custom configuration files as part of the test.

kt-jplotzke: Thanks.  My viewpoint is coming from the standpoint of verifying that the actual functions to read and apply `feature_config.ini` are working correctly during an automated test.  For example, to test if `feature_config.ini` can correctly override the `data_useOccurrence = false` feature, I'd want to put that override line into a configuration file, re-initialize the DLL, and test that functionality.  Otherwise, if you just implemented/used an `EnableFeature` function (which worked correctly), but the base `ReadFeatureFile` function in the DLL had an issue and didn't work correctly, that issue wouldn't be found.

So, in my mind, as long as there are other appropriate tests around the configuration file reading/parsing functionality, the `EnableFeature`/`DisableFeature` could be valid.  However, otherwise I'd opt for actual files.  Just my two cents.

Regardless, automated tests obviously aren't there yet.  I've removed the export of the function and left the `ReloadFeaturesFromFile` as is.

src/path_support.h
kt-jplotzke: Done

src/feature_toggles.h
kt-jplotzke: Done

src/feature_toggles.h
jasonmreding: That's fair. I was speaking primarily from the point of view of wanting for example to run a set of tests with both `data_useOccurrence` enabled and disabled to ensure both code paths work as expected. For that case, I would rather accomplish that by setting the feature flag without having to deploy a config file as part of the test.

jasonmreding:
You may want to revisit the PR description/comments for accuracy in light of the latest changes.

src/path_support.cc
```diff
@@ -0,0 +1,54 @@
...
+            // Fetch the path of the DLL, increasing the buffer size in case the path is a long path
+            std::string path(MAX_PATH, '\0');
+            DWORD result = GetModuleFileNameA(hModule, path.data(), path.size());
+            while (result >= path.size() && GetLastError() == ERROR_INSUFFICIENT_BUFFER) {
+                path.resize(path.size() * 2);
+                result = GetModuleFileNameA(hModule, path.data(), path.size());
+            }
+            path.resize(result); // remove padding at end
+            if (result) {
+                fullPath = std::filesystem::path(path).parent_path();
```
bkeryan, line 38: FYI, I think it would be better to throw an exception if we fail to get the path, but given https://github.com/ni/grpc-labview/issues/437 and the fact that the caller ignores errors from opening the ifstream, I think the current behavior is okay for now. I'll add a note to #437.

bkeryan (approved):
The C++ changes look fine. Thanks for the contribution!

kt-jplotzke:
@jasonmreding - do you have experience with the failing x64/x86 build error?  Seems unrelated to these changes and something to do with the Github runner or build script, but I'd _prefer_ to see a passing build before we merge these changes  :)

```
visualstudio2022-workload-vctools v1.0.0 [Approved] - Possibly broken
visualstudio2022-workload-vctools package files install completed. Performing other installation steps.
Installing visualstudio2022-workload-vctools...

WARNING: visualstudio2022-workload-vctools has been installed. However, a reboot is required to finalize the installation.
WARNING: visualstudio2022-workload-vctools has been installed. However, a reboot is required to finalize the installation.
  visualstudio2022-workload-vctools may be able to be automatically uninstalled.
 The install of visualstudio2022-workload-vctools was successful.
  Software install location not explicitly set, it could be in package or
  default install location of installer.

Chocolatey installed 6/6 packages. 
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Installed:
 - chocolatey-dotnetfx.extension v1.0.1
 - chocolatey-visualstudio.extension v1.11.1
 - dotnetfx v4.8.0.20220524
 - visualstudio2022buildtools v117.14.0
 - visualstudio2022-workload-vctools v1.0.0
 - visualstudio-installer v2.0.3

Packages requiring reboot:
 - visualstudio2022-workload-vctools (exit code 3010)

The recent package changes indicate a reboot is necessary.
 Please reboot at your earliest convenience.
```

kt-jplotzke:
> @jasonmreding - do you have experience with the failing x64/x86 build error?

I've been able to repro this on the `master` branch and have a proposed fix.  I'll issue in a separate PR for discussion.

PR #449, opened 2025-05-21
Set LinuxRT toolchain to 2023Q1

kt-jplotzke:
In order to increase the compatibility of the grpc-labview library on LinuxRT, I am proposing moving from building with the 2023Q4 toolchain (implemented with PR #432) to the 2023Q1 toolchain.  With the current 2023Q4 toolchain, this limits compatibility of the current LinuxRT shared library with NILRT 2023Q4+.  If built with the 2023Q1 toolchain, this increases the compatibility back to NILRT 2022Q4+. 

This toolchain was likely updated due to the grpc v1.62 package only building under gcc 7.3+ (reference build requirements of grpc 1.62 [here](https://github.com/grpc/grpc/tree/f78a54c5ad4e058734aa9b2beb9459940e4de342/src/cpp).  The former NILRT 2018 toolchain only supported gcc 6, and thus needed to be updated.  The 2023Q1 toolchain supports gcc 10.3, while the 2023Q4 toolchain supports gcc 11.3.

However, the 2023Q1 toolchain contains libc 2.33 and libstdc++ 10.3 (which are also compatible with NILRT 2022Q4).
The 2023Q4 toolchain contains libc 2.35 and libstdc++ 11.3, which are only available on NILRT 2023Q4+ and causes the compatibility issues.

You can also reference [Google's Open Source C++ support matrix](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md), but I do not see any incompatibility with the 2023Q1 toolchain.

Further, as the grpc-labview library is intended for LabVIEW 2019+, restricting to NILRT 2023Q4+ doesn't allow for LabVIEW RT 2019 development (as 2023Q4 only has support for LV 2020+).

For testing, I have built the shared library using the 2023Q1 toolchain and tested both client and server functions using NILRT 2022Q4 and all looks functional.
Reference successful build here: https://github.com/kt-jplotzke/grpc-labview/actions/runs/15144256743/job/42575931912

### DIFF
--- a/.github/workflows/build_on_rt.yml
+++ b/.github/workflows/build_on_rt.yml
@@ -26,9 +26,9 @@ jobs:
     # improvements to hosting and exports of the NILRT toolchain are made.
     - name: Install NI Linux RT CC Toolchain
       run: |
-        wget -nv https://download.ni.com/support/softlib/labview/labview_rt/2023Q4/LinuxToolchains/linux/oecore-x86_64-core2-64-toolchain-10.0.sh
-        sudo chmod a+x ./oecore-x86_64-core2-64-toolchain-10.0.sh
-        sudo ./oecore-x86_64-core2-64-toolchain-10.0.sh -y -d ${GITHUB_WORKSPACE}/nilrt-toolchain/
+        wget -nv https://download.ni.com/support/softlib/labview/labview_rt/2018/Linux%20Toolchains/linux/oecore-x86_64-core2-64-toolchain-6.0.sh
+        sudo chmod a+x ./oecore-x86_64-core2-64-toolchain-6.0.sh
+        sudo ./oecore-x86_64-core2-64-toolchain-6.0.sh -y -d ${GITHUB_WORKSPACE}/nilrt-toolchain/
         echo "${GITHUB_WORKSPACE}/nilrt-toolchain/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux" >> ${GITHUB_PATH}
 
     - name: Update Submodules
--- a/.github/workflows/build_on_rt.yml
+++ b/.github/workflows/build_on_rt.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   build-nilrt:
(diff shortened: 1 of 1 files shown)


jasonmreding:
Could you also update the [Readme section](https://github.com/ni/grpc-labview/blob/master/README.md#minimum-compatible-labview-version) discussing LabVIEW compatibility while you are here.

jasonmreding (approved):

jasonmreding:
> @kt-jplotzke it seems something as part of this change is causing the completion of the Linux RT cross compile to not be reported correctly such that the final `build_release_artifacts` stage is triggered. I thought your latest push might resolve it, but it seems something is still broken. I'm not sure how much time I'll have to investigate this independently in the short term. If you need more timely response to anything, feel free to @ mention me as I'm more likely to react to it sooner.

I think I figured it out. There was a branch protection rule that also had to be updated since you renamed the job created by the github action. I'm rerunning things now. Hopefully this will resolve it.

kt-jplotzke:
> > @kt-jplotzke it seems something as part of this change is causing the completion of the Linux RT cross compile to not be reported correctly such that the final `build_release_artifacts` stage is triggered. I thought your latest push might resolve it, but it seems something is still broken. I'm not sure how much time I'll have to investigate this independently in the short term. If you need more timely response to anything, feel free to @ mention me as I'm more likely to react to it sooner.
> 
> I think I figured it out. There was a branch protection rule that also had to be updated since you renamed the job created by the github action. I'm rerunning things now. Hopefully this will resolve it.

@jasonmreding  Hmm, thanks.  Odd since I don't think anything should be renamed.  I also can't repro it using a runner in my forked repo.  Also odd is that my other PR failed the same `build_release_artifacts` step last week when it sat in the "queued" state for 24 hours, and that PR didn't touch any workflow/build actions.

I obviously can't see the self-hosted runner which does that final action.  I see some Github issues of others talking about self-hosted runners sitting in queued states forever and need a reboot to kick them back off.  If this build fails, can you check on that runner and perhaps try a reboot?

PR #451, opened 2025-05-22
Minor usability updates to gRPC code generator

### LINKED ISSUE #447: Generated Enum Flat/Rich converters should not call DLL in UI thread
Currently, the generated VIs for Enum flat/rich conversions use a DLL call to `GetLVEnumValueFromProtoValue` and `GetProtoValueFromLVEnumValue`, but both CLFNs are set to run in UI thread.  These calls should be set to run in any thread for efficiency.

[AB#3133758](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3133758)

### LINKED ISSUE #450: Generated server lvlib should not make methods private that are needed to start the service
When using the latest v1.5.1.1 generation tool (I'm not sure how far back this was implemented), the generated server lvlib makes several top-level VIs in private scope.  In my opinion, this limits useability and workflow of the toolkit.  Specifically, `Register gRPC Messages.vi`, `Register Descriptors.vi`, and `Get Service Class Instances.vi` should be public scope.

For my use case, I have a set of gRPC services which I generate server/client VIs for, but I keep each service as a separate generated lvlib.  I have several versions of a top-level gRPC server which pulls in different gRPC service lvlibs as necessary.  Because these generated VIs are now in private scope, I have to manually reset these to public scope to use these methods outside the lvlib.

I personally only use the generated `Run Service.vi` for individual testing purposes.  I love that it is there, but we shouldn't limit running the generated services to only that VI.

Image

[AB#3135155](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3135155)

kt-jplotzke:
This pull request has several minor "quality of life" improvements to the client/server generator:

- VIs required to run the server are moved to the generated server lvlib public scope.  This allows incorporation of the gRPC service into a larger application.  (Fixes #450)
- Added icon text to "Rich to Flat" and "Flat to Rich" VIs (for both client and server generated VIs)
image

- The generated enum converter VIs ("LV enum to proto enum" and vise-versa) used a CLFN to `labview_grpc_server.dll` to convert LV enum values to/from proto enum values.  As all other server DLL calls are wrapped in `grpc-lvsupport.lvlib` (and also that `TranslateGrpcError.vi` is in private scope of that library), I created two wrapper functions in grpc-lvsupport.lvlib for these calls.  The client template VIs are subsequently updated to use these wrapper VIs instead of the direct CLFN.  (Fixes #447)
- Removed the name for the new local application context that's created by the main generation application.  When using this generator as a subVI (to call it automatically for multiple services), calling it on the 2nd time resulted in an error that the application context name already exists.  Not passing any name to this function ensures that a unique name is created every time, allowing it to run multiple times programmatically.
image

- Updated label for "Generate" control to "Generate Server or Client?" for better naming when using generator tool as a subVI.
image


labview source/Client Server Support New/Server Template/Service Template.lvlib
jasonmreding, line 1: Making these VIs public still seems like a leak of implementation details to me. What if instead we created a new public VI in the template called `Create Services.vi` that was at the same level as `Run Service.vi`? This has better symmetry with `Servicer.lvclass:Create Servier.vi`. It would look something like:

image

and `Run Service.vi` would then look like:

image

Perhaps we should also rename `Run Service.vi` to `Run Server.vi` for consistency. Finding good naming/terminology to clearly distinguish service or API endpoint instances from the hosting application/server has always been a bit tricky. 

Long term, I would like the generator to be more opinionated about what it generates. That would include dictating the library names for the services and either placing each service in its own library or all services from the same protobuf package into a single library. This goes for imported protos as well. It would also be nice if the generated metadata and descriptors were more modular to each service rather than one for the entire library. Not to mention generate a server class that was extended via virtual overrides rather than modifying the generated template code. This is all probably a long way off still unless somebody has a lot of spare time to work on it. However, for now, incremental improvements are appreciated.

labview source/gRPC lv Support/Server API/Server/Register Reflection Descriptors.vi
jasonmreding, line 1: Since you have added new public APIs which are being utilized by the generated code, it would probably be a good idea to update the current package versions for the .vipb files and oldest compatible version for the generator package to correspond to this new version. I would probably update the version to either 1.5.2 or 1.6.1, I don't feel strongly about which is used.

jasonmreding (changes requested):

labview source/Client Server Support New/Server Template/Service Template.lvlib
kt-jplotzke: I like your proposed solution and will update.  I would only recommend the name `Register Services.vi` as the public-facing VI (as opposed to `Create Services.vi`.  I feel like this better aligns with gRPC nomenclature in other languages (where you typically _create_ a server and then _register_ services to it).

For long term, are you open to having a Github discussion thread to throw ideas around for larger changes?  That way, at least smaller incremental updates could be more aligned with towards longer-term goals.

jasonmreding: >  I would only recommend the name Register Services.vi as the public-facing VI (as opposed to Create Services.vi).

I think that would be reasonable as well. I floated `Create Services.vi` since this is the thing that would actually produce a class wire for the service instance, but I guess there is nothing stopping anyone from dropping each generated class instance directly. I don't really see a reason to decouple creation of the services from registration of the services and calling the VI `Create and Register Services.vi` seemed a little verbose, but I could be convinced otherwise on these points. 

> Upon further thought, I agree with your `Create Services.vi` naming as this VI is creating the service objects. I think down the road, I'd rather have a Register Services.vi method in the gRPC Server class that I can pass in an array of gRPC Service objects...

I think you could still call it `Register Services.vi` if you wanted, I think it is about a coin flip for me as to which is better. In this case, the VI would be scoped to the library which owns the generated service class. Later, we would add a public instance method to the service class itself as well as an instance method to `Servicer.lvclass`. All of these VIs would have very similar names which might be a bit confusing, but since they all belong to separate libraries it should at least work.

> For long term, are you open to having a Github discussion thread to throw ideas around for larger changes?

Yes, I am open to suggestions on better long term collaboration, especially if there are common practices across the github community that can be applied here. I know there are some past discussions which have been captured as issues with inconsistent tagging. This is mostly from my own reading as I was not contributing to the repo at the time of those discussions. I don't know if there was ever anything definitive that was decided upon in any of those discussions that could be extracted into a document to serve as a long term vision or design for the repo. You are welcome to read through the open issues for yourself for context. I have my own ideas around the code generation, but I have never tried to document anything in spec form and post it as a RFC.

kt-jplotzke:
When wrapping and creating the `Create Services.vi`, I noticed how much discrepancy there for the FP naming of the servicer/service classes.  Currently, there's a mix of `iServicer`, `OOP Server`, `SignalGeneratorService`, and `TemplateServer`.  I decided to clean all of these up and have consistent naming.  I recognize that this may start straying outside of "minor useability updates", so if you don't like where this is going, I'm happy to revert back the last few commits and only focus on the original intent.

Overall, these changes were made in addition to what's listed in the initial PR.
- `gRPC-servicer.lvlib`
  - `Servicer.lvclass`
    - Standardized all lvclass input/outputs to be named `gRPC server in/out`
    - lvclass icon text already set to orange "SERVER"
  - `ServiceBase.lvclass`
    - Standardized all lvclass input/outputs to be named `gRPC service in/out`
    - Updated lvclass icon text to blue "SERVICE"  (was blue "SERVER")
    - Updated to make sure all VI icons had icon text
- Generator
  - `ServiceName Template.lvclass` (this inherits from `ServiceBase.lvclass`)
    - Standardized all lvclass input/outputs to be named `gRPC service in/out`
    - Updated lvclass icon text to blue "SERVICE"  (was blue "SERVER")
  - `Service Template.lvlib`
    - Standardized all lvclass input/outputs to be named `gRPC service in/out`
    - Updated lvclass icon text to green "SERVICE"  (was green "SERVER")
    
This way, I think there's better distinction on where VIs are coming from at a glance.
image

There were a few places that the generator was calling out the old naming (ie, `OOP Server`).  I went through and updated all references as appropriate and tested several different client/server examples.

labview source/Client Server Support New/Oneof Template/RichToFlat.vim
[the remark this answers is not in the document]
kt-jplotzke: Done

jasonmreding (approved):

jasonmreding:
> @jasonmreding - I see the `build_release_artifacts` step still failing. Did I cause this when I updated the vipb files which seems to inadvertently updated the `Package_LabVIEW_Version` tag from `19.0` to `19.0 (64-bit)`.
> 
> https://github.com/ni/grpc-labview/blob/53c83479ee508230856ab2fed33f1365db755748/labview%20source/gRPC%20lv%20Support/build%20spec/LabVIEW%20gRPC%20Library.vipb#L11
> 
> Should I revert back my VIPB updates and then manually edit the file to update the version and dependencies?

I suspect that is it. I know the self hosted runner only has LV 2019 32 bit installed. IIRC VIPM will automatically modify the LV version for the package if you don't have the version of LV it specifies installed. I think VIPM displays a warning message about this though it may not be obvious what the impact of that message is. You should just be able to edit the xml in a text editor to revert the LV version back to its original form.

jasonmreding:
> > @jasonmreding - Argh, still failing - this time for `There was an error recalling the build specification (invalid build signature).`
> > I opened each of the VIPB files in VIPM, built each package to dry-run the build, then re-saved. Then opened each VIPB file and manually edited the LV version back to 32-bit.
> > Hoping this one helps. Appreciate the patience. :)
> 
> Ok, happened again with same error. Odd since I'm able to run the `BuildGRPCPackages.vi` on my machine (although with LV 2019 64-bit) and it runs ok and generates all packages. My only guess is that opening up the VIPB files changed a GUID/ID in the file that's different between 32-bit and 64-bit and therefore VIPM is complaining on the build machine because something doesn't line up?
> 
> I can try installing LV 2019 32-bit and see if I can repro this issue later.

Yeah, I'm not sure what is going on. The error information returned isn't all that helpful and there isn't much logging available to determine how far things progressed or even which package is reporting the problem. You could try reverting all changes to the .vipb files back to the original other than the package and package dependency version updates and see if that helps. The VIPM error mentions `Source:: BDB9067948AA5DEB4C77B2EF2E3D67FD` but I have no idea what that GUID is referring to.

kt-jplotzke:
> ...
> So I don't know what is going on. Either there is a problem/bug with the different VIPM versions interoperating with each other or there is some weird caching/state issue with VIPM on the self hosted runner machine. As I was poking into this, the runner seems to have gone offline again so this might be a holding pattern until somebody can diagnose that again.

Thanks.  I've also reached out to JKI to see if they can shed any light for what an "invalid build signature" is and if they'd expect those GUIDs/ID values that change when I re-save the file to represent something specific or are they just random values that we can ignore.

kt-jplotzke:
@jasonmreding  - I reverted back the VIPB files, and then opened them on a system with LV 2019 32-bit and set all version properties again.  VIPM 2021 can now open these files.

I think the root cause is that these files can't be edited manually and the "ID" field must be a checksum of sort.  In order to update these files, one must have LV 2019 32-bit installed so that VIPM doesn't change that field.

BShermanator:
In addition to the "RegistergRPC Messages.vI" reverting back to public, there are other accessors and type definitions that need to be moved back to public. Outlined the changes in this issue: https://github.com/ni/grpc-labview/issues/453

kt-jplotzke:
> In addition to the "RegistergRPC Messages.vI" reverting back to public, there are other accessors and type definitions that need to be moved back to public. Outlined the changes in this issue: #453

@jasonmreding  - How about as a compromise, we move all of these items back to public scope, but issue a deprecation notice with the release notes that these will eventually be moved back protected?  I do feel like some users moved up to v1.5.0.1 for the security updates (myself included) and then were surprised/not expecting the protected-scope items. 

With this, v1.6.0.1 maintains backwards-compatibility with previous applications which use these VIs.  We still leave the new `Create Services.vi` in place and use that in the generated example `Run Server.vi` and advise that that is the intended approach.

Then, once the base issues that cause people to use these VIs (like grabbing the UEs out of the class) are resolved, a v2.0.0 release is targeted which moves these back to protected-scope and issues a warning for breaking changes.

jasonmreding:
@kt-jplotzke The `Register gRPC Methods.vi` on `ServiceBase.lvclass` is still marked as protected which is breaking the generated server hierarchy. Also, I believe other changes will be made as a result of renaming the generated `Run Service.vi` to `Run Server.vi`. If you grep for that string, you will see some hits to it in Python test scripts. There are also some string constants to it in some LV test VI wrappers. If you want to revert that change and submit it in a follow on PR, I am fine with that. I'm also OK if you want to keep moving forward with the change. Just wanted to warn you it might take a few more iterations to find/fix all of the fall out from that. We should probably update some of the repo documentation as well because of it as a number of tutorials reference that VI by name.

kt-jplotzke:
> @kt-jplotzke The `Register gRPC Methods.vi` on `ServiceBase.lvclass` is still marked as protected which is breaking the generated server hierarchy. Also, I believe other changes will be made as a result of renaming the generated `Run Service.vi` to `Run Server.vi`. If you grep for that string, you will see some hits to it in Python test scripts. There are also some string constants to it in some LV test VI wrappers. If you want to revert that change and submit it in a follow on PR, I am fine with that. I'm also OK if you want to keep moving forward with the change. Just wanted to warn you it might take a few more iterations to find/fix all of the fall out from that. We should probably update some of the repo documentation as well because of it as a number of tutorials reference that VI by name.

@jasonmreding - Updates:
- Set `Register gRPC Methods.vi` in `ServiceBase.lvclass` and the `ServiceName Template.lvclass` back to protected scope to align with previous generated code.  This shouldn't be called by the user and was inadvertently set when `Register gRPC Messages.vi` should be the VI in public scope.
-Updated `New_ATS` code to call `Run Server.vi` instead of `Run Service.vi`.  I got the ATS code running and am able to run all tests on my machine with passing results.
-`AutoTests` were previously failing because of the mixed protected/public scope on `Register gRPC Methods.vi`.  I am now able to run all AutoTests on my machine with passing results.
-Updated docs to use `Run Server.vi` instead of `Run Service.vi`

Note that the examples still use the old nomenclature and would need to be re-generated to align with the documentation.  This may affect the AutoTests execution.  Do you want that as part of this PR?

jasonmreding:
> Note that the examples still use the old nomenclature and would need to be re-generated to align with the documentation. This may affect the AutoTests execution. Do you want that as part of this PR?

I think this PR has already sprawled on longer than what would be considered ideal. At this point, I would rather perform additional refactoring/cleanup in follow on PRs that are easier to review in isolation.

PR #482, opened 2025-12-03
 Update LabVIEW build workflow to support multiple versions of LabVIEW: Try 2

Faraclas:
#Purpose 
LabVIEW 2019 is currently needed for maintaining or contributing to this project. However, requiring LabVIEW 2019 to build in general was an unnecessary restriction.  Requiring this presented a potential barrier to those either wanting to get started contributing but not wanting to commit to installing an antiquated LabVIEW IDE, or those wanting to fork the project an use or maintain in a later version of LabVIEW.  Now any version (2019 or later) can be used to build.

#Changes 
The majority of the changes involved finding all of the hardcoded '2019' version strings in the code and reworking these into inputs.  This can be seen here in the top level.  Of course the version also needed to be passed in from python, along with port information for to support multiple versions coexisting and listening.  The changes were made in build.py to support this and the documentation adjusted accordingly,

image

One issue that should be noted can be seen in the creation of the package below.  The .vipb files could only be used with the specific labview version that they were saved in using JKIs toolchain.  The workflow implemented overcomes this issue and enables this work to be successful.

image

Otherwise, code was cleaned up in general to follow recommended guidelines when possible:
* CTRL+F to position front panels and block diagrams.
* General cleaning during the edits.
* Execution set for shared-clone reentrant (except for the VI called by LV CLI, that will not get parameters if it is reentrant)

#Testing
- Using this code I built the VIPM packages using LabVIEW 2019, 2023, and 2025.
- Example commands used to build:
```
python .\build-it\build.py --target Win64 --labview-version 2019 --labview-bits 32
python .\build-it\build.py --target Win64 --labview-version 2023 --labview-bits 64
```


build-it/LV Build/BuildGRPCPackages.vi
jasonmreding: You are building up an array using conditional tunnels and then indexing the array afterwards to grab the first element. However, you immediately terminate the loop as soon as you find the match so there will only ever be one element. You can simplify things by just using a last value tunnel mode and deleting index array.

Image

jasonmreding: The `labview version` input appears to be wired to a bottom terminal on the connector pane. However, it is really connected to a terminal on the top of the connector pane. I would recommend either:
- Relayout/wire so it's clear where the connection is
- Move the terminal to the bottom of the connector pane for each sub VI
- Chain the version in/out and from left to right similar to the error terminals

I kind of prefer the last option, but I can live with any of them.

Image

jasonmreding: There is also a `Switch Target.vi` on the VIPM palette which looks like it might be able to be used to change the default LV version when nothing is wired. However, the documentation isn't entirely clear and I haven't tested it. If it does work, it might eliminate a lot of tedious changes to chain this information through all the APIs.

Image

build-it/LV Build/EditVIPBLVVersion.vi
jasonmreding: nit: Cleanup wiring a bit and avoid wires that go behind nodes. I had to move some of the wiring around in order to clearly see what was wired to what terminals.

PR #483, opened 2025-12-08
Sign windows DLLs with NI License before building vipm packages.

yash-ni:
Sign windows DLLs before making the VIP packages


nischalks:
This is now fixed in https://github.com/ni/grpc-labview/releases/tag/v1.6.1.1

PR #484, opened 2025-12-09
Update Reentrancy in LabVIEW Source

Faraclas:
Purpose: 
* While most of the source VIs were reentrant, some of them were not resulting in less than optimum performance for generated labview-grpc APIs.
* In general, this code was comprised of mostly preallocated clones that are marked to be in-lined.  As such, attempts were made to follow this when possible.  There were a few instances where shared clones were used because of dynamic dispatch terminals or other situations which may have called for this (top level API etc).

Changes Made: The following lists the VIs that were changed and what type of reentrancy they were set to:
grpc-labview\labview source\gRPC lv Support
- Shared\To Well Known Type.vi (preallocated / inline)
- Shared\To Message Name.vi (preallocated / inline)
- Shared\Is Message Well Known (preallocated / inline)
- Server API\Cluster to gRPC message\Element to Message Element Metadata.vi (preallocated / inline)
- \Server API\Server\Register Reflection Descriptors.vi (preallocated)
- \Server API\Server\Get Server DLL Path.vi (shared clone, can't in-line a VI with the contents of this one.)

grpc-labview\labview source\gRPC lv Servicer
- ServiceBase\Accessors\Read gRPCid.vi
- ServiceBase\Accessors\Read Server Stop.vi (shared clone, dynamic dispatch)
- ServiceBase\Server API\Destroy.vi (shared clone, dynamic dispatch)
- ServiceBase\Server API\Instantiate.vi (shared clone, dynamic dispatch)
- ServiceBase\Server API\Register gRPC Methods.vi (shared clone, dynamic dispatch)
- Servicer\Run Servicer.vi (shared clone)
- Servicer\Accessors\Server State\Get Server State.vi (preallocated / inline. Since its not dynamic dispatch, accessors are best in-lined)
- Servicer\Accessors\Server State\Set Server State.vi (preallocated / inline. Since its not dynamic dispatch, accessors are best in-lined)
- Servicer\Server API\Create Server.vi (shared clone)
- Servicer\Server API\Initialize Server State.vi (shared clone)
- Servicer\Server API\Start Server.vi (shared clone)
- Servicer\Server API\Stop Server.vi (shared clone)

Client Server Support New\Client Template
- RPC Messages\Get package_template_enum for Value.vi (preallocated / inline)
- RPC Messages\Get Value for package_template_enum.vi (preallocated / inline)

Client Server Support New\Oneof Template
- GetActiveFieldName.vi (preallocated / inline)
- HasTemplate.vi (preallocated / inline)

Client Server Support New\Server Template
- Cleanup Panel.vi (shared clone)
- Create Services.vi (shared clone)
- Get Service Class Instances.vi (shared clone)
- Initialize Panel.vi (shared clone)
- Monitor for Stop.vi (shared clone)
- Run Server.vi (preallocated)
- Generated Descriptors\Generated Descriptors.vi (preallocated / inline)
- RPC Messages\Get package_template_enum for Value.vi (preallocated / inline)
- RPC Messages\Get package_template_enum.vi (preallocated / inline)
- RPC Messages\Get package_templatemessage.vi (preallocated / inline)
- RPC Messages\Get Value for package_template_enum.vi (preallocated / inline)
- RPC Messages\Register Descriptors.vi (preallocated)
- RPC Messages\Register gRPC Messages.vi (preallocated / inline)
- RPC Messages\Set package_template_enum.vi (preallocated / inline)
- RPC Messages\Set package_templatemessage.vi (preallocated / inline)
- RPC Service\ServiceName\RPC Methods\Bind RPC package_service_method.vi (preallocated)
- RPC Service\ServiceName\RPC Methods\Get package_service_method Request.vi (preallocated / inline)
- RPC Service\ServiceName\RPC Methods\RPC Method package_service_method.vi (preallocated)
- RPC Service\ServiceName\RPC Methods\Set package_service_method Response.vi (preallocated / inline)
- RPC Service\ServiceName\Server API\Destroy.vi (shared clone)
- RPC Service\ServiceName\Server API\Register gRPC Methods.vi (shared clone)

Testing: The VIPM packages were built using the same command the CI will be using.


labview source/Client Server Support New/Oneof Template/HasTemplate.vi
jasonmreding: Why did you add an in place element structure here?

jasonmreding (changes requested):
I'm curious what motivated this change and what process you used to identify the VIs that were changed. While I think most of these changes are probably harmless, I suspect it will actually have very little impact on performance. On the other hand, some of these changes will degrade the debugging experience for the server side code. 

The most important thing to ensure is reentrant on the server side is any code related to RPC dispatching/handling. That applies mostly to some of the template VIs you changed here. Even then, I doubt this will make much difference as we generate VIs per RPC so there really won't be any conflicts over parallel execution anyway. All of the code called as part of initialization of the server really doesn't need to be reentrant at all as there shouldn't be any parallelism involved there. To keep the debug experience as pleasant as possible, I would prefer to revert the reentrancy changes to those VIs.

labview source/Client Server Support New/Oneof Template/HasTemplate.vi
Faraclas: This VI was already marked as preallocated/in-line.  The only reason to do that is for performance.  While I was in that VI, I figured I would improve the performance by adding the in-line element structure.

jasonmreding (approved):

jplotzke (approved):

Arrowin-87:
Isn't it much simpler to let this design choice (performance versus debug experience) up to the user using a checkbox in the code generator which sets the reentrancy by enabling this checkbox ON/OFF?
Preferably (speaking for myself), I would not be interested in debugging the generated code as I am expected it will work anyway.. In this case I would be pleased to make them re-entrant if I have a need for multiple gRPC server and/or client setups..

jasonmreding:
> Isn't it much simpler to let this design choice (performance versus debug experience) up to the user using a checkbox in the code generator which sets the reentrancy by enabling this checkbox ON/OFF? Preferably (speaking for myself), I would not be interested in debugging the generated code as I am expected it will work anyway.. In this case I would be pleased to make them re-entrant if I have a need for multiple gRPC server and/or client setups..

Most of the VIs changed in this PR are VIs that get deployed to vilib so they would not be impacted by any reentrancy configuration options added to the code generation of each client/server.

PR #490, opened 2026-01-07
Add ClientOpenClientStreamingCall

Faraclas:
# Pull Request: Add ClientOpenClientStreamingCall for Immediate Server Events

## Summary

This PR adds a new function `ClientOpenClientStreamingCall` that allows the server to receive an event immediately when a client streaming call is opened, before any data is written. This enables more responsive server-side handling of streaming connections.

## Motivation

With the existing `ClientBeginClientStreamingCall`, the server doesn't receive an event until the client writes the first message. This delay can be problematic when:
- The server needs to perform setup operations before receiving data
- The server needs to acknowledge the connection immediately
- The application requires low-latency connection awareness

### New Function: `ClientOpenClientStreamingCall`

Added a new client-side function with the same signature as `ClientBeginClientStreamingCall`:

```cpp
int32_t ClientOpenClientStreamingCall(
    grpc_labview::gRPCid* clientId,
    const char* methodName,
    const char* requestMessageName,
    const char* responseMessageName,
    grpc_labview::gRPCid** callId,
    int32_t timeoutMs,
    grpc_labview::gRPCid* contextId)
```

### Behavior Comparison

| Aspect | ClientBeginClientStreamingCall | ClientOpenClientStreamingCall |
|--------|-------------------------------|------------------------------|
| Event timing | After first write | Immediately on stream open |
| Initial event data | Contains first message | Empty (no data) |
| Use case | Standard streaming | Connection-aware streaming |

### Implementation Details

**Client-side (`src/grpc_client.cc`, `src/grpc_client.h`):**
- New `ClientOpenClientStreamingCall()` function that sets custom metadata
- Adds metadata header `x-grpc-labview-stream-mode: open` to signal open stream mode
- Sets `_isOpenStream` flag on `ClientStreamingClientCall` class
- Tracks `_initialEventSent` state

**Server-side (`src/event_data.cc`, `src/grpc_server.h`):**
- Detects "open stream" metadata in incoming calls
- Immediately raises event to LabVIEW when detected (before reading data)
- Sets `_initialEventRaised` flag to track state
- Skips normal Process flow that waits for first message

**Interop layer (`src/grpc_interop.cc`):**
- Added `GetCallStatusCode()` method to `CallData` class
- Returns custom status codes set via `SetCallStatus` for better error handling

### Files Modified

- `src/grpc_client.h` - Added flags to `ClientStreamingClientCall` class
- `src/grpc_client.cc` - Implemented `ClientOpenClientStreamingCall()` function
- `src/grpc_server.h` - Added flags to `CallData` class and `GetCallStatusCode()` method
- `src/event_data.cc` - Modified event handling logic to support immediate events
- `src/grpc_interop.cc` - Added status code checking in `GetRequestData()`

**Statistics:** 5 files changed, 135 insertions(+), 4 deletions(-)

## Backward Compatibility

✅ **Fully backward compatible** - All existing code using `ClientBeginClientStreamingCall` continues to work unchanged. The new function is additive only.

## Protocol

The implementation uses gRPC metadata to communicate between client and server:

| Metadata Key | Value | Purpose |
|--------------|-------|---------|
| `x-grpc-labview-stream-mode` | `"open"` | Signals server to raise event immediately |

## Usage Example (Conceptual)

**Before (ClientBeginClientStreamingCall):**
```
Client: Open stream → Write message 1
Server:                                  ← Receive event with message 1
```

**After (ClientOpenClientStreamingCall):**
```
Client: Open stream
Server:              ← Receive event (no data yet)
Client:              → Write message 1
Server:              ← Can read message 1 when ready
```

## Testing

- ✅ Built and verified on Windows
- ✅ Function properly exported from DLL
- ✅ Backward compatibility verified with existing code
- ✅ Metadata protocol tested end-to-end

### DIFF
--- a/src/grpc_interop.cc
+++ b/src/grpc_interop.cc
@@ -419,6 +419,11 @@ LIBRARY_EXPORT int32_t SetResponseData(grpc_labview::gRPCid** id, int8_t* lvRequ
 
         if (!data->_call->IsActive() || !data->_call->Write())
         {
+            auto statusCode = data->_call->GetCallStatusCode();
+            if (statusCode != grpc::StatusCode::OK)
+            {
+                return -(1000 + statusCode);
+            }
             return -2;
         }
         return 0;


jasonmreding:
I have posted #491 which was inspired by this change. I think it is a simpler and cleaner change that I believe to be safe without having to introduce any one off or special cases in the code. Assuming things looks good with that, I would propose we abandon this PR in favor of that change.

PR #491, opened 2026-01-08
Signal LV events for the server when connection for RPC is established

jasonmreding:
# Description
Signal LV events for the server when connection for RPC is established rather than waiting until the first request message is read. This provides more flexibility in the server implementation for a few different scenarios:
- The server needs to perform setup operations before receiving data
- The server needs to acknowledge the connection immediately
- The application requires low-latency connection awareness
- A bidirectional RPC wants to send response data to the client before receiving any request data

# Implementation / Design
- LV event for the RPC is now signaled when the connection for the RPC is established rather than when the first request message is received. Request messages are now only read from the stream when explicit API calls are made from the user's implementation of the RPC on the server. We no longer read the first request message in response to events from the server's completion queue.
- Updated some of the internal state machine state names for readability.
- Deleted `CallData::ReadComplete`. This is no longer needed as all reads are now performed within `ReadNext`.

# Testing
I manually tested using unary, client streaming, server streaming, and bidirectional RPCs using the example route guide service. I also spot tested using some existing measurement services.

### DIFF
--- a/src/event_data.cc
+++ b/src/event_data.cc
@@ -8,17 +8,16 @@
 using namespace google::protobuf::internal;
 
 namespace grpc_labview
-{  
+{
     //---------------------------------------------------------------------
     //---------------------------------------------------------------------
     CallData::CallData(LabVIEWgRPCServer* server, grpc::AsyncGenericService *service, grpc::ServerCompletionQueue *cq) :
-        _server(server), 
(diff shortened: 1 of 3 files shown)


irwand:
This seems like a big ish change from the client perspective? should it be 1.6.2 release?

Issue #494, opened 2026-02-03
Not able to regenerate client code after changing .proto file


rknobel:
Hello. 
I am trying to follow your Quick Start, and I am getting an error after I added one more method to the server in the .proto file.
I can only regenerate code if I change the library name. Then, it creates two libraries in the same project. For example helloword and helloworld1. 

Error message:
Error 1026 occurred at Property Node in TRef Traverse.vi->VI Scripting - Traverse.lvlib:Traverse for GObjects.vi->gRPC Template Creation Utility-release.lvlib:Get ControlRef from ctrlVI.vi->gRPC Template Creation Utility-release.lvlib:Generate Descriptor VI.vi->gRPC Template Creation Utility-release.lvlib:Main.vi (Traverse Failed)

LabVIEW: (Hex 0x402) VI Reference is invalid.

<img width="334" height="278" alt="Image" src="https://github.com/user-attachments/assets/6dc7c497-2c0e-4a20-8c48-986e76a9fb89" />

Issue #480, opened 2025-11-13
Server fails to rebuild if adding new rpc to .proto file


Mad3DScientist:
Create a server using the following .proto file

```
syntax = "proto3";
package ni_hal;
//==========================================================================================
// Message Types
//==========================================================================================
message null {}

message mBool {
  bool value = 1;
}

message mDouble {
  double value = 1;
}

//===========================================================================================
//  GRPC SERVICE DEFINITION
//===========================================================================================

// Unless otherwise indicated, distances are in microns, time in milliseconds

service SlowService {
  rpc GetLocation(null) returns (mDouble);               // return current location of motion
}

service FastService {
  rpc Abort(null) returns (mDouble);                     // immediately stop all motion, may require reset afterwards
}
```

Replace the SlowService in the .proto file with the following

```
service SlowService {
  rpc GetLocation(null) returns (mDouble);               // return current location of motion
  rpc GetSpace(null) returns (mDouble);
}

```

Regenerate the server. It will error during creation and no code will be updated.

[AB#3491856](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3491856)

PR #509, opened 2026-05-22
Partial Solution to Server Regeneration When Modifying .proto File

Madd3DScientist:
- Purpose: During development of a gRPC server, it is common to add, subtract, and modify commands. It is expected that modifying the `.proto` file then regenerating the server would work. It does not. Currently, the programmer must generate a differently named server, copy the code from the event structures from the old to the new, delete the old and rename the new. As servers get larger, this gets more difficult and it is easy to make mistakes.
- Changes Made: This change introduces a script which automates this copying. The included README explains the process and lists conditions and known bugs. Total of two files added, the script and its README.
- Related Issues: #494, #480
- Testing: Take an existing server, modify its .proto file, and follow the directions in the README. This script has been tested with both large and small projects. It will fail if the restrictions in the README are not followed.  

Ideally, the server creation script would scan the project for current elements and modify if present, rather than failing. However, writing that script would take awhile. Thus, this workaround.


jasonmreding:
@Madd3DScientist I posted #513 which I'm hoping eliminates the need to submit these types of scripting/tooling work arounds. If you have time to look at and test these changes against the services you have been working with, I would be interested to hear your feedback. In particular, I'm curious whether that PR mostly addresses the needs that caused you to write this script or what other pain points remain where you would still keep this script around.

Madd3DScientist:
I should get a chance to check it this week. Thanks!

PR #510, opened 2026-05-22
RT Quickstart Updates

Madd3DScientist:
- **Purpose**: This addresses two pain points new users of gRPC-LabVIEW deploying on LV RT run into. These two pain points are copying the `.so` file to the target and using the `Run Server.vi` unmodified.
- **Changes Made**: A Quickstart README is added for RT users. This could alternately be rolled into the main Quickstart README. The `Run Server.vi` template is modified with conditional compile statements to remove the front panel references which cause it to fail. This could alternately become a `Run Server RT.vi`.
- **Related Issues**: none documented
- **Testing**: Verify the locations in the Quickstart RT documentation. Then create and deploy a server to an RT target and run `Run Server.vi`. It should execute without error.


labview source/Client Server Support New/Server Template/Run Server.vi
jasonmreding, line 1: For `Initialize Panel.vi` and `Cleanup Panel.vi`, there are already conditional disable structures inside the VIs. For these VIs, I would prefer adding new cases for RT to those structures rather than wrapping the VIs themselves in new structures.

For the other two conditional disable structures which were added, I think it would be a little cleaner and more readable to just have single structure with all of the platform specific code combined rather than two different structures. I realize that means you might need to fork the error wire to `Start Server.vi` further up stream. I think that is fine and shouldn't matter functionally. In fact, I'm not sure how the current VI's data flow doesn't block startup when executing under RT? It doesn't seem like the `Start Server.vi` can execute until the `Stop` button is pressed. Unfortunately, I don't have an RT target available where I can test any of this.

Madd3DScientist: I have made the requested changes, but am a remote worker, so will be unable to test until Monday, when I am in the office and have access to an RT target. I will let you know. Thank you for the feedback!

jasonmreding: > I have made the requested changes, but am a remote worker, so will be unable to test until Monday, when I am in the office and have access to an RT target. I will let you know. Thank you for the feedback!

I don't see any new commits pushed after the RP was posted. Did you forget to push or did you determine no additional changes were required? If you haven't had time to revisit this, that is fine. I just want to make sure we weren't both waiting on the other to respond.

Madd3DScientist: I have not pushed them yet. Our IT department made some network security changes which killed our ability to talk to RT targets, so I have been unable to test. I should get this done by Monday. I will add a reply here when the changes are pushed.

Madd3DScientist: I finally got a working RT system for test. Still need to set everything up, but I should get it done by noon tomorrow.

Madd3DScientist: Just pushed it. I had to wrap the entire `Init Panel.vi` code in a conditional compile, as the query for front panel state was triggering an error. Also found the .so file will not work on older ARM RT devices (I used a cRIO-9066), which I expected to happen. You get an error 7 from the call library node.

Seems to be ready to go.

jasonmreding: @Madd3DScientist FYI, the test pipeline [failed](https://github.com/ni/grpc-labview/actions/runs/27354520412/job/80849013353?pr=510). Not sure if you get github notifications for that or not since an owner has to approve the workflow to run on your behalf since the tests execute on a self hosted runner. 

Anyway, the logs through the LabVIEW CLI provide minimal information so you will probably want to try and reproduce the problem locally. If you are not familiar with that test pipeline, you can find some information [here](https://github.com/ni/grpc-labview/tree/master/tests/New_ATS) for how to setup and execute the tests locally.

Madd3DScientist: Found a reference change I missed and fixed it. Should work now.

jasonmreding (approved):

PR #512, opened 2026-05-27
Adding client.vi to manually test grpc feature

Danus-NI:
Modification done to fix Main_CLIWrapper:
Added `To word integer`:
image

Added client.vi to manually run and test the grpc labview client - labview server connection.
Screentshops of the vis:
Helloworld:
image
image

all-datatypes-oneof:
image
image

enum-oneof:
image
image

message-oneof:
image
image

multiplie-field-oneof:
image
image

nested-oneof:
image
image

sibline-oneof:
image
image

simple-oneof:
image
image

enum-streaming:
1. clientstreaming
image
image

2.server streaming:
image
image

oneof-streaming:
1. client streaming
image
image

2. server streaming
image
image


tests/New_ATS/pylib/testlist.json
```diff
@@ -13,7 +13,7 @@
       "oneof-streaming",
       "enum-streaming"
     ],
-    "gen_type": "0",
+    "gen_type": "2",
```
yash-ni: I think for now we should revert this to only generate server until we automate LabVIEW client testing.

tests/New_ATS/Main_CLIWrapper.vi
jasonmreding, line 1: I don't understand why the changes to this VI were necessary. I think the file should probably be reverted.

tests/New_ATS/pylib/testlist.json
Danus-NI: Alright will revert it back to generate server only

tests/New_ATS/Main_CLIWrapper.vi
Danus-NI: Hi Jason,
The Main_CLIWrapper was not responding according when the "gen_type" was set to 2 (generate both client and server folders) for all the cases it was only generating server folder, so I have added a case structure to the VI to fix this issue.
I will be adding the screenshot of the same in the PR right away for better clarification

jasonmreding: This is the original VI. Are you saying the `Type Cast` was failing and it was always producing the `gRPC Server` enum value?

image

jasonmreding: Well... I just tried a quick test, and it does fail to convert properly. It appears the integer type must match the enums underlying type for the type cast to work. I don't know if LV supports a format string that produces an I16 rather than an I32. If not, you can always use `To Word Integer` type cast before casting to the enum type.

Danus-NI: I have used the `To Word Integer` now, and the issue seems to be resolved
image

Danus-NI: @jasonmreding  Will be marking this issue as resolved

yash-ni (approved):

jasonmreding (approved):
From a code style perspective:
- The diagram wiring is a bit messy in spots. Try to maintain parallel wires without bends where possible. When not possible, try to maintain the minimal number of wire bends for a given wire. For example, try to avoid this sort of thing. 

Image

- Do not use the icon view for controls/indicators on the block diagram. I recommend you uncheck the "Place front panel terminals as icons" in the options dialog.

Image

The remainder of the client VIs are a bit difficult to provide feedback on yet as nothing references them yet. For example, I suspect having all of the client VIs in the same folder as the server without any consistent naming scheme will be inconvenient when you attempt to modify the test suite scripts to copy them into the generated code in a generic fashion. However, I guess some of this feedback can be deferred until those changes come along.

jasonmreding:
BTW, these changes seem to be causing the build_release_artifacts pipeline stage to timeout/fail. I haven't tried to investigate why.

Danus-NI:
I have updated the messy wiring section as suggested:
image
@jasonmreding

PR #517, opened 2026-06-12
Tests failures should fail pipeline stage

jasonmreding:
The `build_release_artifacts` workflow has a `Run CI Tests` step that can never fail because the python script does not return a non-zero exit code when test failures are detected.


tests/run_tests.py
```diff
@@ -27,9 +27,10 @@ def run_all_tests():
             run_result = run_test(test_runner_vi, os.path.join(test_directory, test))
             # run_result will be None for PASSED tests
             if(run_result != None):
-                failed_test_results += run_result 
+                failed_test_results += run_result
         if (failed_test_results != ""):
-            _logger.exception(failed_test_results)
```
bkeryan, line 32: Ok, that makes sense. _logger.exception takes a str but it expects to be called from an except block.
https://docs.python.org/3/library/logging.html#logging.Logger.exception

bkeryan (approved):

examples/helloworld/helloworld_client/RPC Messages/helloworld_HelloReply.ctl
bkeryan, line 1: @jasonmreding I think you probably need to merge main into this branch

